PR #2, opened 2019-08-26
See in commit descriptions

joclefrancois:


gertsonderby:
An immediate issue with this pull request is that it changes two things. The changes to DateInput should be in their own PR - though I've already solved the issue in question (see [this commit](https://github.com/Orckestra/orc-shared/commit/d96e4ef66ab875b47bf9810a0ec756574eeed9f0)).

The changes made to Scope etc. are something I want to take a closer look at, so I have a clear idea what's going on there.

src/components/Treeview/Label.js
```diff
@@ -1,7 +1,7 @@
 import styled from "styled-components";
 import Icon from "../Icon";
 import { getThemeProp, ifFlag } from "../../utils";
-import { branchHeight } from "./settings";
+import { branchLength } from "./settings";
```
gertsonderby: Why make this change? Why not instead set the default value (and value in theme) to a different number?

src/components/Treeview/Label.js
```diff
@@ -21,15 +21,23 @@ export const Indicator = styled(Icon).attrs({
...
 `;
 
+export const BeforeIndicator = styled.div`
+	background-color: #666;
+	margin-left: -${props => branchLength(props) + 1}px;
+	width: ${props => branchLength(props) + 1}px;
+	height: 1px;
+	z-index: 99;
+	position: absolute;
+`;
```
gertsonderby: I'd like an explanation of what this element does, and why it is necessary. As far as I can tell it's a horizontal line, so overlaps in function with NonIndicator below it. They're also used together. Why can't it be a single element?

gertsonderby:
> usually here, we don't bother to submit a quick fix in the same PR of another story, especially when it is minor, 

Having unrelated matters in a PR is bad practice, no matter how small it is. PRs should be as small and focused as possible, and not contain unrelated changes. Not all PRs need to have a user story, either, and in specific bug fixes like this one should not have one - if anything they tie to bug reports. And they should *absolutely* not be hidden with PRs that do other things.

gertsonderby:
Another note on this PR: If the title of the PR just refers to the commit messages, and there is no description, the commit messages need to be highly informative. Please be more descriptive in future.

src/components/Treeview/Label.js
gertsonderby: I checked the code out and went over it, this seems fine, overall.

src/components/Treeview/Label.js
gertsonderby: All right, this checks out.

PR #3, opened 2019-08-30
Category list

gertsonderby:
Adds a CategoryList component which shows rows as members of closable groups.


gertsonderby:
Needs documentation.

PR #9, opened 2020-02-17
Reimplement many HOCs and components using hooks

gertsonderby:
Deprecation warnings have been added in both code and documentation for HOCs which have been replaced.

No API changes have been made. Some HOCs will no longer provide props in excess of what the documentation says - typically routing properties - but as these props were undocumented and an unintended side effect, this is not considered an API change.


src/components/List/List.js
```diff
@@ -67,19 +68,34 @@ const calculateVirtualization = (
...
+};
+
 export const List = ({
+	name,
 	columnDefs = [],
 	rows = [],
 	rowOnClick,
 	placeholder,
-	selection = [],
+	// selection = [],
```
gertsonderby: Should be removed.

src/components/Loader.js
```diff
@@ -1,24 +1,41 @@
...
 import LoadingIcon from "./LoadingIcon";
 import ErrorPlaceholder from "./ErrorPlaceholder";
 
-export const Loading = ({ error, retry, pastDelay }) => {
+export const Loading = ({ error }) => {
+	const [pastDelay, flagDelay] = useState(false);
+	let timer = useRef(null);
+	useEffect(() => {
+		// Prevent refiring delay timeout
+		if (!timer.current) {
```
gertsonderby: Should also check if delay flag already set.

PR #10, opened 2020-03-11
The current scope was not always visible when opening the selector.

joclefrancois:
If the parent scope state was closed, the scope tree was not expanded to see the current scope.

Story: #26826

### DIFF
--- a/src/components/Scope/index.js
+++ b/src/components/Scope/index.js
@@ -6,6 +6,7 @@ import useViewState from "../../hooks/useViewState";
 import useScopeData from "./useScopeData";
 import Button from "../Button";
 import Selector from "./Selector";
+import usePreviousModified from "../../hooks/usePreviousModified";
 
 export const Bar = styled.div`
 	flex: 0 0 49px;
@@ -41,6 +42,11 @@ export const Scope = ({ children, filterPlaceholder }) => {
 	const name = "scopeSelector";
 	const [currentScope, defaultNodeState, getScope] = useScopeData();
 	const [{ show = false, nodeState, filter }, updateViewState] = useViewState(name);
+	const opening = usePreviousModified(
+		show,
+		(prevShow, currentShow) => prevShow === false && currentShow === true,
+	);
+
 	const reset = event => {
 		updateViewState("show", false);
 		event.stopPropagation();
@@ -61,7 +67,7 @@ export const Scope = ({ children, filterPlaceholder }) => {
 				reset={reset}
 				getScope={getScope}
 				nodeState={nodeState}
-				defaultNodeState={defaultNodeState}
+				defaultNodeState={opening ? defaultNodeState : {}}
 				filter={filter}
 				updateFilter={updateFilter}
 				filterPlaceholder={filterPlaceholder}
--- a/src/components/Scope/index.js
+++ b/src/components/Scope/index.js
@@ -42,9 +42,10 @@ export const Scope = ({ children, filterPlaceholder }) => {
 	const name = "scopeSelector";
 	const [currentScope, defaultNodeState, getScope] = useScopeData();
 	const [{ show = false, nodeState, filter }, updateViewState] = useViewState(name);
-	const opening = usePreviousModified(
+	usePreviousModified(
 		show,
-		(prevShow, currentShow) => prevShow === false && currentShow === true,
+		current =>
+			current && updateViewState("nodeState", { ...nodeState, ...defaultNodeState }),
 	);
 
 	const reset = event => {
@@ -52,6 +53,7 @@ export const Scope = ({ children, filterPlaceholder }) => {
 		event.stopPropagation();
 	};
 	const updateFilter = event => updateViewState("filter", event.target.value);
+
 	return (
 		<React.Fragment>
 			<ScopeBar
@@ -66,8 +68,6 @@ export const Scope = ({ children, filterPlaceholder }) => {
 				show={show}
 				reset={reset}
 				getScope={getScope}
-				nodeState={nodeState}
-				defaultNodeState={opening ? defaultNodeState : {}}
 				filter={filter}
 				updateFilter={updateFilter}
 				filterPlaceholder={filterPlaceholder}
--- a/src/components/Scope/index.js
+++ b/src/components/Scope/index.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import styled from "styled-components";
 import { Route, Switch, Redirect } from "react-router-dom";
 import { getThemeProp } from "../../utils";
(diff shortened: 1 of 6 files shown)


src/hooks/usePreviousModified.js
```diff
@@ -0,0 +1,15 @@
...
+const usePreviousModified = (value, predicate) => {
+	const ref = useRef();
+
+	useEffect(() => {
+		ref.current = value;
+	}, [value]);
+
+	return predicate
+		? predicate(ref.current, value)
+		: ref.current !== undefined && ref.current !== null && ref.current !== value;
```
gertsonderby: Better done as a default value for `predicate`, see `useLoader` and its `cutoutSelector` parameter for inspiration.

src/components/Treeview/index.js
```diff
@@ -20,7 +20,7 @@ export const Treeview = ({
 	...otherProps
 }) => {
 	const [viewState, updateViewState] = useViewState(name);
-	const nodeState = { ...defaultNodeState, ...viewState.nodeState };
+	const nodeState = { ...viewState.nodeState, ...defaultNodeState };
```
gertsonderby: This is the opposite of what makes sense for the component. Viewstate is intended to override the default state, by design, so that user modifications are changes to it.

This is not a change that can be accepted into the library, as a result.

src/components/Treeview/Treeview.test.js
```diff
@@ -233,7 +233,7 @@ describe("TreeView", () => {
 		);
 	});
 
-	it("overrides default node state with view state", () => {
+	it("overrides view state with default node state", () => {
```
gertsonderby: See comments on source file.

src/components/Scope/index.js
```diff
@@ -61,7 +67,7 @@ export const Scope = ({ children, filterPlaceholder }) => {
 				reset={reset}
 				getScope={getScope}
 				nodeState={nodeState}
-				defaultNodeState={defaultNodeState}
+				defaultNodeState={opening ? defaultNodeState : {}}
```
gertsonderby: This is better done as a modification or filter on the `nodeState` field. The scope selector maintains state between unmount and mount through this - as a feature of the TreeView component. If you were to set an effect on it where you reset the nodeState field when unmounting (cf. [`useEffect` cleanup](https://reactjs.org/docs/hooks-effect.html#effects-with-cleanup)), for example, that would drop the persistent state and go back to the default. Other options exist for solving this in different ways.

src/components/Scope/index.js
```diff
@@ -41,11 +42,18 @@ export const Scope = ({ children, filterPlaceholder }) => {
 	const name = "scopeSelector";
 	const [currentScope, defaultNodeState, getScope] = useScopeData();
 	const [{ show = false, nodeState, filter }, updateViewState] = useViewState(name);
+	usePreviousModified(
+		show,
+		current =>
+			current && updateViewState("nodeState", { ...nodeState, ...defaultNodeState }),
```
gertsonderby: You end up missing some dependencies for useEffect here. If defaultNodeState changes, the effect function here won't be updated - `usePreviousModified` will keep using the one it was passed with at first. This means there's potential for ugly bugs down the line.

I'd advise a useEffect hook here which depends on both show, nodeState and defaultNodeState, and resets the nodestate if show is false. Something like:
```js
useEffect(() => {
  if (show) updateViewState(<as above>);
}, [show, nodeState, defaultNodeState]);
```

joclefrancois: Well, after some investigation, it is a "no can't do" unfortunately. Calling "updateViewState" within the useEffect would update the state over and over (as you know already I'm sure) and using the nodeState object in the array of the second parameter not prevent the useEffect to be triggered as it will be different that the previous time everytime we call "updateViewState". It does not perform a deep object comparison. Here are two urls about that issue and some suggestions are made but I am not sure we want to go that path as for now. I am trying to find another solution but I would like your feedback first.

https://stackoverflow.com/questions/55808749/use-object-in-useeffect-2nd-param-without-having-to-stringify-it-to-json/55808960?noredirect=1#comment98705815_55808960

https://stackoverflow.com/questions/53601931/custom-useeffect-second-argument

gertsonderby: I think it could be made to work. I'll dig into it further in the morning exactly how. I do think your hook may work for it, but it needs the right effect function, potentially a useCallback-wrapped one. (But it's evening here now, so I'm not at my very sharpest. :relaxed:)

joclefrancois: Oh, I have no doubt there is a solution and I also thought about a callback method. Let's talk about it tomorrow morning. I was not expecting an answer from you so soon btw. Have a good evening.

gertsonderby: Okay, here's what I came up with: Set the default node state on the TreeView component as it was before. When `viewState.show` turns true, set `viewState.nodeState`  to {}. This deletes the previous node state, and the default node state takes over.

I'll add some comments in the code diff to clear this up.

src/components/Scope/index.js
```diff
@@ -60,8 +68,6 @@ export const Scope = ({ children, filterPlaceholder }) => {
 				show={show}
 				reset={reset}
 				getScope={getScope}
-				nodeState={nodeState}
-				defaultNodeState={defaultNodeState}
```
gertsonderby: Reinstate this - that way TreeView handles the default. (The nodeState prop is not received by Selector, so that can stay gone.)

src/components/Scope/index.js
```diff
@@ -41,11 +42,18 @@ export const Scope = ({ children, filterPlaceholder }) => {
 	const name = "scopeSelector";
 	const [currentScope, defaultNodeState, getScope] = useScopeData();
 	const [{ show = false, nodeState, filter }, updateViewState] = useViewState(name);
+	usePreviousModified(
+		show,
+		current =>
```
gertsonderby: This function can be simplifed to `current => current && updateViewState('nodeState', {})`.

src/components/Scope/index.js
gertsonderby: There. This also means you don't need the current `viewstate.nodeState`, since all you do it reset it.

joclefrancois: Hi Gert, I understand perfectly what you are saying, however, like I told you last week, I feel kind of sad that we lose the feature to have expanded branches somewhere else in the scope tree if I set viewState.nodeState to {}... just saying. Allow me to do one more suggestion before we come to that please. See my next push before to comment (unit tests will be fixed later, after comments and/or modifications). Basically, I only call the action in my hook when the predicate returns true, that is, when the selector is opening.

Now, IMHO, if a set nodeState to {} everytime we open the selector, I don't see the purposes anymore to keep that state in the REDUX store. I have the feeling that useState() would be enough for the Scope component. Again, don't get me wrong please, I am not trying to argue here, I am only trying to state how I understand things, whether I am right or wrong. If the latter, I would like to know your reason to understand better.

gertsonderby: I'll resolve this conversation, awaiting your update.

gertsonderby:
Published to npm in v0.8.18-dev.3

PR #17, opened 2020-04-08
Bugs/36423 title and chevron not visible enough

joclefrancois:
Based on our UX/UI expert, the new base color was not dark enough for other components in the OMS but it is still fine for the TopBar components. I also had to change this library to reflect those requirements. Also, the drop down menu chevron was a bit too small to his taste.

### DIFF
--- a/src/components/DropMenu/Anchor.js
+++ b/src/components/DropMenu/Anchor.js
@@ -4,17 +4,20 @@ import { getThemeProp, ifFlag } from "../../utils";
 import Icon from "../Icon";
 
 export const Header = styled.div`
+	display: flex;
 	cursor: pointer;
 	color: ${ifFlag(
 		"open",
-		getThemeProp(["colors", "application", "base"], "#ccc"),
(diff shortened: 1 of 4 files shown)


src/components/DropMenu/Anchor.js
```diff
@@ -4,38 +4,43 @@ import { getThemeProp, ifFlag } from "../../utils";
...
 	)};
 `;
 
 const Anchor = ({ id, onClick, className, menuLabel, open }) => (
 	<Header {...{ id, onClick, className, open }}>
 		{menuLabel}
-		<Indicator open={open} />
+		<IndicatorWrapper>
+			<Indicator open={open} />
+		</IndicatorWrapper>
```
gertsonderby, line 43: This wrapper serves no apparent purpose - I advise removing it. Never add extraneous elements to the DOM tree.

joclefrancois: Oh well, I had added for styling reason originally, but now, it looks without it. Removed.

PR #18, opened 2020-04-09
Sign in back was not working correctly with the return URL.

joclefrancois:
window.location.origin was not working as expected unfortunately. I had to use window.BASE_PATH in my function instead of "basename" because otherwise, I would not be able to unit tests the sign-out properly.


joclefrancois:
@gertsonderby, well, I'm sorry but I'm am quite note sure why Travis is complaining. It does not seem to be related to my changes. Hope you can help me.

PR #19, opened 2020-04-13
Added the required Help button at the top-right corner

joclefrancois:
- Added the Help button at the top-right corner of the shell
- Added the call the a new request in order to fetch the version and Help URLs (OcrkestraCommerce PR is still pending, should be reviewed by Benoit tuesday)
- Both OMS and Analytics will need to be changed in order to:
1- Add the localisations of the Help button
2- The new Prop when using AppFrame
3- The call to the new action "getVersionInfo()" in index.js

### DIFF
--- a/docs/infrastructure.md
+++ b/docs/infrastructure.md
@@ -38,7 +38,7 @@ The `utils` folder contains a number of useful utility functions, for use in var
 
 ### API utilities
 
-`loadConfig()` loads in the `config.json` file (by fetching it at runtime, not as an imported module) in order to gain access to the API location, version number etc. It returns a promise, which when resolved signals the availability the API. This is required to be called first and allowed to complete before dispatching API actions using the `buildUrl()` helper function. <!-- Edit this to generalize if/when we have more API helper funcs
+`loadConfig()` loads in the `config.json` file (by fetching it at runtime, not as an imported module) in order to gain access to the API location, version number etc. It returns a promise, which when resolved signals the availability the API. This is required to be called first and allowed to complete before dispatching API actions using the `buildUrl()` helper function. <!-- Edit this to generalize if/when we have more API helper funcs  -->
 
 `buildUrl(pathParts, queryObject)` constructs an absolute URL for an Orckestra API end point. The `pathParts` parameter is an array of strings defining the path, for example `['my', 'application']` will target the `/api/my/application` endpoint. `queryObject` should be an object of key/value pairs that will become search query parameters, for example `{ IncludeChildren: true, IncludeCurrency: true }` turns into `?IncludeChildren=true&IncludeCurrency=true`. It is best used together with [`makeOrcApiAction()`](actionsreducersselectors.md#api-action-tools), q.v.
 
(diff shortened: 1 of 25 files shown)


docs/infrastructure.md
```diff
@@ -38,7 +38,7 @@ The `utils` folder contains a number of useful utility functions, for use in var
 
 ### API utilities
 
-`loadConfig()` loads in the `config.json` file (by fetching it at runtime, not as an imported module) in order to gain access to the API location, version number etc. It returns a promise, which when resolved signals the availability the API. This is required to be called first and allowed to complete before dispatching API actions using the `buildUrl()` helper function. <!-- Edit this to generalize if/when we have more API helper funcs  --> It also sets `window.orcVersion` to the current Orckestra Commerce Suite version number.
+`loadConfig()` loads in the `config.json` file (by fetching it at runtime, not as an imported module) in order to gain access to the API location, version number etc. It returns a promise, which when resolved signals the availability the API. This is required to be called first and allowed to complete before dispatching API actions using the `buildUrl()` helper function. <!-- Edit this to generalize if/when we have more API helper funcs
```
gertsonderby, line 42: Yikes, you have accidentally commented out half the doc file here!

src/actions/versionInfo.js
```diff
@@ -0,0 +1,14 @@
...
+export const GET_VERSION_INFO = "GET_VERSION_INFO";
+
+export const [
+	GET_VERSION_INFO_REQUEST,
+	GET_VERSION_INFO_SUCCESS,
+	GET_VERSION_INFO_FAILURE,
+] = makeActionTypes(GET_VERSION_INFO);
+
+export const getVersionInfo = () =>
+	makeOrcApiAction(GET_VERSION_INFO, buildUrl(["diagnostic/versioninfo"]));
```
gertsonderby, line 14: This should be in array form, not a slashed url.

src/components/AppFrame/useVersionInfo.js
```diff
@@ -0,0 +1,18 @@
+import { useSelector } from "react-redux";
```
gertsonderby, line 1: This hook should probably be renamed to something like `useApplicationHelpUrl` - since that's what it does. Always keep your names sensible, otherwise you will create problems for yourself and others down the line. And better a 200-character function name than one that doesn't tell you what it does!

docs/infrastructure.md
joclefrancois: Yikes indeed, sorry for the misstep.

src/components/AppFrame/useVersionInfo.js
joclefrancois: You're right... you know, I am also an experienced developer (mostly back-end though) and I fully agree that names should be appropriate. Still, naming things has always been a huge deal in my daily tasks :laughing:. Sometimes here in Montreal, we discuss about the right name for several minutes before to agree on something.

I admit that on this one, I did not pay much attention as I used basically the name as my actions, selectors & reducers, with that said, I like your suggestion and I will use it.

gertsonderby:
I   guess I was tired, missing that this in fact is an API change. Well, good thing we haven't made a v1.0 of orc-shared yet!

PR #23, opened 2020-04-21
Changed how languages options should be handled in preferences.

joclefrancois:
Languages options in Preference panel should be limited to those three choices only down below as per product owner requirements. The request to fetch supported cultures has other purposes completely and should not be used in preferences.

- English
- Français
- Italiano

"locales" in package.json should be built this way so user may switch from one language to another.

```
"locales": [
    {"language": "English", "cultureIso": "en-US"},
    {"language": "Français", "cultureIso": "fr-CA"},
    {"language": "Italiano", "cultureIso": "it-IT"}
]

```

### DIFF
--- a/src/buildStore.js
+++ b/src/buildStore.js
@@ -42,7 +42,9 @@ const buildStore = (reducers, devOptions = {}) => {
 		applyMiddleware(routerMiddleware(history), apiMiddleware, spawnerMiddleware),
 	);
 
-	const supportedLocales = SUPPORTED_LOCALES || ["en"];
+	const supportedLocales = SUPPORTED_LOCALES || [
+		{ language: "English", cultureIso: "en-US" },
+	];
 	const localeReducer = localeFactory(supportedLocales);
 
 	buildReducer = reducers =>
--- a/src/reducers/localeFactory.js
+++ b/src/reducers/localeFactory.js
@@ -2,17 +2,20 @@ import Immutable from "immutable";
 import { CHANGE_LOCALE, GET_CULTURES_SUCCESS } from "../actions/locale";
 
 const localeFactory = supportedLocales => {
+	const defaultCulture =
+		(supportedLocales[0] && supportedLocales[0].cultureIso) || "en-US";
+
 	const initialState = Immutable.fromJS({
-		locale: supportedLocales[0],
+		locale: defaultCulture,
 		supportedLocales,
 		cultures: {},
-		defaultCulture: supportedLocales[0],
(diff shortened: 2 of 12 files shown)


src/reducers/localeFactory.js
```diff
@@ -2,17 +2,20 @@ import Immutable from "immutable";
 import { CHANGE_LOCALE, GET_CULTURES_SUCCESS } from "../actions/locale";
 
 const localeFactory = supportedLocales => {
```
neexite, line 4: Documentation actionsreducersselectors.md is not updated (Now it 'An array of IETF language tags')

src/reducers/localeFactory.js
```diff
@@ -2,17 +2,20 @@ import Immutable from "immutable";
...
+	const defaultCulture =
```
neexite, line 5: @gertsonderby Do we have some app that uses 'supportedLocales' in the old format(array of tags) ?

src/buildStore.js
```diff
@@ -42,7 +42,9 @@ const buildStore = (reducers, devOptions = {}) => {
 		applyMiddleware(routerMiddleware(history), apiMiddleware, spawnerMiddleware),
 	);
 
-	const supportedLocales = SUPPORTED_LOCALES || ["en"];
+	const supportedLocales = SUPPORTED_LOCALES || [
+		{ language: "English", cultureIso: "en-US" },
```
neexite, line 46: Do we need to change 'en' to 'en-US'? In other places it still 'en' as default value ('src/reducers/localeFactory.js' , 'src/selectors/locale.js')

src/reducers/localeFactory.js
joclefrancois: Analytics and OMS will need to be changed accordingly when updating to the next version of orc-shared. This is in my to do list for OMS. Analytics might be handled by Gert himself.

gertsonderby: This change needs to be made when updating the app to use the versions of orc-scripts and orc-shared that contain this.

joclefrancois:
> This will require a change in orc-scripts in order to keep extract-messages working. It currently expects the `locale` field to be an array of strings.

joclefrancois:
@gertsonderby 
I will wait for a new version of orc-scripts before to complete and merge this PR.

gertsonderby:
Latest changes should be in orc-scripts@1.0.0-dev.6. Make sure that is set in both dev and peer dependencies.

PR #27, opened 2020-04-28
Remove recompose HOCs from withClickOutside

gertsonderby:
I've reconstructed withClickOutside to no longer use recompose, instead using hooks and refs. This does mean it will only work on components that accept refs, but since it's intended to be used only on DOM elements and styled components etc., that should not cause issues.


src/hocs/withClickOutside.js
```diff
@@ -11,24 +10,21 @@ const descendsFrom = (descendant, ancestor) => {
 	}
 };
 
-const withClickOutside = lifecycle({
-	componentDidMount() {
-		this.innerRef = ReactDOM.findDOMNode(this);
-		this.handler = event => {
-			if (!descendsFrom(event.target, this.innerRef)) {
-				(this.props.onClickOutside || (() => {}))(event);
+const withClickOutside = Comp => ({ onClickOutside = () => {}, ...props }) => {
```
TONYFECTEAU, line 13: Nice!  So, if a component needs to know when a click is outside of it, this hoc can be used.

gertsonderby: Yep, it's been in for a while, I just rebuiilt it to eliminate recompose and its deprecation warnings.

PR #28, opened 2020-04-28
Help URL and web site does not open with the right locale.

joclefrancois:
The help does not open in the user language set up in preferences, neither does the web site when opening it from the about dialog when the user language is in french. For the web site, I had a discussion with Jean-Walter and Benoit and that how they wanted it even though I disagreed, oh well.

Story: #36936


src/components/AppFrame/AppFrame.js
```diff
@@ -57,10 +59,12 @@ const AppFrame = ({
 	scopeFilterPlaceholder,
 	noScope,
 }) => {
+	const locale = useSelector(currentLocale);
 	const applications = unwrapImmutable(useSelector(localizedAppSelector));
 	const [helpUrl] = useApplicationHelpUrl(applicationId);
 	useLoader(getApplications(), state => localizedAppSelector(state).size);
 	const [open, toggle, reset] = useToggle(initOpen);
+	useLoader(getVersionInfo(locale), () => helpUrl !== null);
```
TONYFECTEAU, line 67: If the language is changed, this will not get version info for the new language: /fr

src/selectors/applications.js
```diff
@@ -1,5 +1,5 @@
 import { createSelector } from "reselect";
-import { currentLocale } from "./locale";
```
TONYFECTEAU, line 2: Is currentLocale still available, or will OMS and Analytics need to change?

src/selectors/locale.js
```diff
@@ -13,12 +13,14 @@ const supportedLocales = createSelector(
 	data => data.getIn(["supportedLocales"]) || Immutable.fromJS([]),
 );
 
-export const currentLocale = createSelector(
+export const currentLocaleOrDefault = createSelector(
```
TONYFECTEAU, line 16: This will force all dependant apps to use the new selector name.

joclefrancois: I did not know how to name the new function to return the current locale even when the value was still null. I had preferred to rename the existing one. Update OMS & Analytics is not a big deal anyway.

PR #32, opened 2020-04-29
Removed recompose from Loader component

gertsonderby:


src/components/Loader.test.js
```diff
@@ -5,87 +5,24 @@ import { mount, act } from "unexpected-reaction";
...
 	});
 	afterEach(() => {
 		clock.restore();
 	});
-	spyOnConsole(["error"]);
 
-	it("loads the component", () => {
-		const Comp = Loader(buttonLoader);
+	it("loads and renders the component", () => {
+		const Comp = Loader(compLoader);
```
TONYFECTEAU, line 25: Why the uppercase? Because it is a component?

TONYFECTEAU: Cool!  Clock timers!  I have to learn sinon a little better....

PR #34, opened 2020-04-30
Remove recompose from withInfiniteScroll HOC

gertsonderby:
This one is a bit more complex, and harder to test because jsdom doesn't support scroll events. I'm putting it as a draft for now, and leaving it up to you guys to finish when you're ready for it. It should only need to be tested (and potentially bugfixed) in the browser to ensure that infinite scroll still works well.


joclefrancois:
Ok, we will address that as soon as we can.

joclefrancois:
> This one is a bit more complex, and harder to test because jsdom doesn't support scroll events. I'm putting it as a draft for now, and leaving it up to you guys to finish when you're ready for it. It should only need to be tested (and potentially bugfixed) in the browser to ensure that infinite scroll still works well.

joclefrancois:
I closed it by mistake.

PR #38, opened 2020-05-14
Web site should open in a new tab.

joclefrancois:
Story: #37284

### DIFF
--- a/src/components/AppFrame/Help.test.js
+++ b/src/components/AppFrame/Help.test.js
@@ -13,7 +13,7 @@ jest.mock("../../utils/buildUrl", () => {
 });
 
 describe("Help", () => {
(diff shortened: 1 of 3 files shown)


src/components/AppFrame/Help.test.js
```diff
@@ -36,14 +36,8 @@ describe("Help", () => {
 		modalRoot = document.createElement("div");
 		modalRoot.id = "modal";
 		document.body.appendChild(modalRoot);
-
-		windowOpen = window.open;
```
TONYFECTEAU, line 40: Why the windowOpen?  In case it needs to be accessed?

joclefrancois: Lines were removed, not added.

PR #39, opened 2020-05-14
Help should open in English when any other languages than French.

joclefrancois:
Story: #37283

### DIFF
--- a/src/actions/versionInfo.js
+++ b/src/actions/versionInfo.js
@@ -1,6 +1,7 @@
 import { makeActionTypes } from "./makeApiAction";
 import makeOrcApiAction from "./makeOrcApiAction";
 import { buildUrl } from "../utils/buildUrl";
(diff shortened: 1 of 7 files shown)


src/actions/versionInfo.js
```diff
@@ -15,7 +16,7 @@ export const getVersionInfo = locale =>
 	makeOrcApiAction(
 		GET_VERSION_INFO,
 		buildUrl(["diagnostic", "versioninfo"], {
-			cultureName: locale,
+			cultureName: locale.substr(0, 2).toLowerCase() === "fr" ? locale : cultureByDefault,
```
TONYFECTEAU, line 19: If it is Italian, we will default to English.... I guess this is Ok for now, but Italian will be translated eventually.

joclefrancois: The decision was from Product Owner.

PR #41, opened 2020-05-27
Material ui integration

Crispried:
Material UI was integrated to project. Some wrappers were created for some MUI components.


joclefrancois:
> material

joclefrancois:
> Looks good... except I think you are missing a reference to material UI version in the packages.json?
> I like that your wrapper verifies the proper objects are passed in.

joclefrancois:
@TONYFECTEAU,  We don't add librairies directly in orc-shared, we must do it in orc-script, which was done already 2 weeks ago.

src/components/MaterialUI/Surfaces/ExpansionPanel.js
```diff
@@ -0,0 +1,128 @@
...
+import ExpansionPanelMui from "@material-ui/core/ExpansionPanel";
+import ExpansionPanelSummary from "@material-ui/core/ExpansionPanelSummary";
+import ExpansionPanelDetails from "@material-ui/core/ExpansionPanelDetails";
+import ExpansionPanelActions from "@material-ui/core/ExpansionPanelActions";
+import { makeStyles } from "@material-ui/core/styles";
+import classNames from "classnames";
+import Icon from "./../../Icon";
+import { ExpansionPanelProps, ExpansionPanelActionsProps } from "./expansionPanelProps";
+
+const useStyles = makeStyles(theme => ({
```
TONYFECTEAU, line 11: Is this the best place to put the themes for each control?  I am not saying this is bad... this means if a theme change is to be done, we will be targeting each of the wrappers.

src/components/MaterialUI/Surfaces/ExpansionPanel.test.js
```diff
@@ -0,0 +1,114 @@
...
+	});
+
+	it("Fails if expansionPanelProps has wrong type", () => {
+		var component = <ExpansionPanel />;
+		var mountedComponent = () => {
+			component.props.expansionPanelProps = "Wrong type";
+
+			return mount(component);
+		};
+		expect(mountedComponent, "to throw");
```
TONYFECTEAU, line 37: Is there a way to test what is thrown?

src/components/MaterialUI/Surfaces/SectionsExpansionPanel.test.js
```diff
@@ -0,0 +1,114 @@
...
+	});
+
+	it("Fails if expansionPanelProps has wrong type", () => {
+		var component = <SectionExpansionPanel />;
+		var mountedComponent = () => {
+			component.props.expansionPanelProps = "Wrong type";
+
+			return mount(component);
+		};
+		expect(mountedComponent, "to throw");
```
TONYFECTEAU, line 37: Same comment, can we test the error thrown?

src/components/MaterialUI/Surfaces/ExpansionPanel.test.js
```diff
@@ -0,0 +1,114 @@
...
+import ExpansionPanelMui from "@material-ui/core/ExpansionPanel";
+import ExpansionPanelSummary from "@material-ui/core/ExpansionPanelSummary";
+import ExpansionPanelDetails from "@material-ui/core/ExpansionPanelDetails";
+import ExpansionPanelActions from "@material-ui/core/ExpansionPanelActions";
+import { ExpansionPanelProps, ExpansionPanelActionsProps } from "./expansionPanelProps";
+
+describe("Expansion Panel", () => {
+	it("Renders Expansion Panel", () => {
+		var header = <p>Header</p>;
+		var content = <p>Content</p>;
```
TONYFECTEAU, line 13: Can we test multiple contents?  We will be using that for various components.

src/components/MaterialUI/componentProps.test.js
```diff
@@ -0,0 +1,63 @@
+import ComponentProps from "./componentProps";
+
+describe("Component Props", () => {
+	it("Sets the prop value", () => {
+		var componentProps = new ComponentProps();
```
TONYFECTEAU, line 5: Can be a const....

src/components/MaterialUI/componentProps.test.js
```diff
@@ -0,0 +1,63 @@
...
+
+		componentProps.componentProps.set("key1", null);
+
+		componentProps.set("key1", "value1");
+
+		expect(componentProps.componentProps.get("key1"), "to equal", "value1");
+	});
+
+	it("Throws if prop key is missing", () => {
+		var componentProps = new ComponentProps();
```
TONYFECTEAU, line 15: const, const...

src/components/Provision.js
```diff
@@ -32,26 +33,30 @@ const Fonts = createGlobalStyle`
...
 
-const Provision = ({ store, theme = {}, children }) => (
-	<Provider store={store}>
-		<ConnectedRouter history={history}>
-			<ThemeProvider theme={theme}>
-				<React.Fragment>
-					<Head />
-					<GlobalStyle />
-					<Authenticate>
+const Provision = ({ store, theme = {}, muiTheme, children }) => {
```
TONYFECTEAU, line 36: No test for passing null or undefined for muiTheme....

src/components/MaterialUI/Surfaces/ExpansionPanel.js
Crispried: @TONYFECTEAU 
It's not control it's an ability to use material theme in your styles. so, if you need for example to get color from MUI theme you can access it using that parameter like color: theme.palette.primary.dark. 

theme should have same structure (designed by MUI) so, we can be sure that theme.palette.primary.dark will exist anyway.

src/components/MaterialUI/Surfaces/ExpansionPanel.test.js
Crispried: Yes there is but, I'm throwing just an error. It's possible to throw TypeError for those cases. However I would say it doesn't make sense. The main point here that developer will be able to see in console an error message which will explain him that he passed parameter with wrong type in that wrapper (message is the main tip - not an error type).

src/components/MaterialUI/Surfaces/SectionsExpansionPanel.test.js
Crispried: replied in previous comment

src/components/MaterialUI/Surfaces/ExpansionPanel.test.js
Crispried: What do you mean by multiple content? What wrapper makes is just puts passed header, content and actions in appropriate places. What we tested here is that he actually puts it in necessary places of MUI expansion panel. 

it will put components which are passed and the complexity of those components doesn't make any difference on how it will work.

src/components/MaterialUI/componentProps.test.js
Crispried: why const for a local variable? consts should be used for places where variable shouldn't be overwritten by design. In this case I don't care about it there aren't such requirements.

src/components/MaterialUI/componentProps.test.js
Crispried: why const for a local variable? consts should be used for places where variable shouldn't be overwritten by design. In this case I don't care about it there aren't such requirements.

src/components/Provision.js
Crispried: that kinda useless tests honestly speaking but, I will add it ok..

src/components/MaterialUI/Surfaces/ExpansionPanel.test.js
Crispried: Replaced Error with TypeError and adjust tests with this.

src/components/MaterialUI/Surfaces/SectionsExpansionPanel.test.js
Crispried: Replaced Error with TypeError and adjust tests with this.

src/components/MaterialUI/componentProps.test.js
Crispried: Replaced all the var with let

src/components/MaterialUI/componentProps.test.js
Crispried: Replaced all the var with let

src/components/Provision.js
Crispried: Added test when no mui theme is passed to provision

joclefrancois:
> Material UI was integrated to project. Some wrappers were created for some MUI components.

joclefrancois:
Don't forget to solve issues reported by Travis build.

PR #44, opened 2020-06-04
Added Select Wrapper component from Material-UI and localization helper functions

joclefrancois:
Story: #36654

- Created Select Wrapper from the Material-UI
- Created muiThemes helper file so common themes for all applications (OMS, Analytics, future ones) will be in one place one
- Added helper functions for localization

### DIFF
--- a/src/components/MaterialUI/Inputs/Select.js
+++ b/src/components/MaterialUI/Inputs/Select.js
@@ -61,7 +61,9 @@ const ChevronDown = props => (
 	</div>
 );
 
-const Select = ({ options, update, value }) => {
+const Select = selectProps => {
+	const { value, update, options } = selectProps;
+
 	const classes = useStyles();
 
 	const handleChange = event => {
--- a/src/components/MaterialUI/Inputs/Select.js
+++ b/src/components/MaterialUI/Inputs/Select.js
@@ -1,0 +1,89 @@
+import React from "react";
+import MenuItem from "@material-ui/core/MenuItem";
+import SelectMUI from "@material-ui/core/Select";
+import { makeStyles } from "@material-ui/core/styles";
+
+const useStyles = makeStyles(theme => ({
+	select: {
+		"& ul": {
+			//top: '20px',
+			minWidth: "175px",
+			maxHeight: "300px",
+		},
+		"& li": {
+			fontSize: 12,
+			"&:hover": {
+				backgroundColor: theme.palette.primary.main,
+			},
+		},
+	},
+	input: {
+		borderRadius: 4,
+		position: "relative",
+		backgroundColor: theme.palette.background.paper,
+		border: "1px solid #ced4da",
+		fontSize: 12,
+		minWidth: "150px",
+		maxWidth: "200px",
+		padding: "6px 6px 6px 6px",
+		transition: theme.transitions.create(["border-color", "box-shadow"]),
+		"&:focus": {
+			borderRadius: 4,
+			borderColor: "#80bdff",
+			boxShadow: "0 0 0 0.2rem rgba(0,123,255,.25)",
+		},
+	},
+}));
+
+const MenuProps = {
+	getContentAnchorEl: null,
+	anchorOrigin: {
+		vertical: "bottom",
+		horizontal: "left",
+	},
+};
+
+const ChevronDown = props => (
+	<div
+		{...props}
+		style={{
+			fontSize: "32px",
+			fontFamily: "courier,monospace",
+			//fontWeight: 'bold',
+			color: "black",
+			height: "10px",
+			width: "20px",
+			marginRight: "5px",
+			padding: "12px 0 0",
+		}}
+	>
+		ˇ
+	</div>
+);
+
+const Select = ({ options, update, value }) => {
+	const classes = useStyles();
+
+	const handleChange = event => {
+		update(event.target.value);
+	};
+
+	return (
+		<SelectMUI
+			value={value}
+			onChange={handleChange}
+			disableUnderline={true}
+			IconComponent={ChevronDown}
+			data-test-id={"SelectMUI"}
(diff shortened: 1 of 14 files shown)


src/components/MaterialUI/Inputs/Select.js
```diff
@@ -0,0 +1,88 @@
...
+			width: "20px",
+			marginRight: "5px",
+			padding: "12px 0 0",
+		}}
+	>
+		ˇ
+	</div>
+);
+
+const Select = ({ options, update, value }) => {
```
Crispried, line 64: I think it's better to create selectProps and move options and value there. Sample of that you can find in expansionPanelProps for Expansion Panel. Then we have to pass in wrapper just selectProps object which will include options/value props (and more props probably in future). By this approach we will avoid necessity to rewrite Select component signature (like properties) in future what will allow us to avoid necessity to adjust all the usages of it in OMS.

src/components/MaterialUI/muiThemes.js
```diff
@@ -0,0 +1,467 @@
...
+	muiTheme.props = setThemeProps(muiTheme);
+	muiTheme.overrides = setThemeOverrides(muiTheme);
+
+	return {
+		applicationMuiTheme: applicationMuiTheme,
+		muiTheme: muiTheme,
+	};
+};
+
+export default createThemes;
```
Crispried, line 467: What for you've added mui theme inside orc-shared? Themes should be defined in applications (there is a high chance that in different application theme elements will be different).

src/components/MaterialUI/muiThemes.test.js
```diff
@@ -0,0 +1,69 @@
...
+
+		expect(muiTheme.direction, "to equal", "ltr");
+		expect(muiTheme.secondary.main, "to equal", "#ED2E0B");
+
+		expect(
+			muiTheme.overrides.MuiButton.outlined.border,
+			"to equal",
+			"1px solid rgba(255, 255, 255, 0.23)",
+		);
+	});
```
Crispried, line 68: I don't think that it make sense to write unit tests on theme as I don't think it make sense to write tests on any styles at all. Argument is pretty simple - theme values cannot be an independent unit. I mean all those values exists only as part of context of other components/parts of application. As style cannot be an independent unit as well. Values for styles depends on context and correctness of those values is really relative.

src/utils/propertyHelper.js
```diff
@@ -0,0 +1,28 @@
+export function getPropertyOrDefault(
```
Crispried, line 1: Maybe it's too late for me but, I don't understand the use case for this helper function and I wasn't able to find where did you use it. I don't have any concerns about it but, just would like to understand the use case for it.

src/components/MaterialUI/muiThemes.js
joclefrancois: See my comments, I think there will be common themes between applications. If something needs to be different, then, you will be able to do it customize in the application. I see no points to have same overrides in each our projetcts.

src/components/MaterialUI/muiThemes.test.js
joclefrancois: That brought test coverage at 100%, it does not hurt to have a test for that, does it?

src/utils/propertyHelper.js
joclefrancois: getPropertyOrDefault is used in localizationHelper, this is partly code we have been using in MSF and it does what we need to accomplished for localisations as we expect and as I told you yesterday. I will need getLocalization in the OMS and it will be useful throughout all our applications.

joclefrancois: It is just a helper function that allows us to get a property from an object with some use cases we had in MSF, eg, we needed to ignore case or we wanted to look for a property starting with some characters. Maybe it does sound silly for you but it has been useful in MSF in some use cases.

src/components/MaterialUI/muiThemes.js
joclefrancois: If you want me to rename that file, I can do it, but I will not copy/paste the OMS file into Analytics as such as it will ask double maintenance everytime we need to add overrides or change typography for all of our applications. In the future, we will have four applications using orc-shared, so every time we will need to change the overrides, we will need to do it in four projects. I did it only to ease maintenance in the future.

joclefrancois: Maybe I can named it "commonThemes" or "basicThemes", not sure

src/components/MaterialUI/Inputs/Select.js
joclefrancois: I change the props to have only one, however, I don't see the point to have it strong typed as 99% of the code is not anyway in orc-shared.

Crispried: The way how I've implemented componentProps allows final user to use only material ui docs and intellisense without going to orc-shared and checking which keys he have to use for properties and how they will be used in wrapper. So, the point is usability and unification.

src/components/MaterialUI/muiThemes.js
Crispried: but, how you are expecting to use it? the point is that currently theme is passed from oms to orc-shared. and because of that MUI theme could be accessed in makeStyle hook in both solutions. However if we won't pass theme from app to orc-shared (let's say app didn't define it's own theme) it means that we won't be able to use theme from that App. or what it will be necessary to do is to directly define MUI theme for that app as Theme imported from orc-shared.

Anyway I'm strictly not agree that we have to define "default theme". the default theme is already defined by MUI just to enable possibility to use "fallback values". But, different application will have lot of differences in colors fonts in which global classes colors/fonts/sizes should be used etc. That's what I got from discussion with Alain and that's why we are defining theme inside application and not inside orc-shared

PR #45, opened 2020-06-08
Features/37019 shipment metadata

Anstsiya:
wrapper for text to have ability to control number of lines for wrapping and truncate on the last line


joclefrancois:
We have been trying to keep code coverage at 100%, if it is simple to bring it up to that level, would it be possible to add necessary unit tests ? Run "npm run coverage" locally, or look at Travis build, and it will give you line number that are not covered.

src/components/MaterialUI/NLinesText.js
```diff
@@ -0,0 +1,68 @@
...
+			},
+		},
+		lineClampWebkit: {
+			WebkitLineClamp: props.lineCount,
+			display: "-webkit-box",
+			WebkitBoxOrient: "vertical",
+		},
+	}));
+
+const NLinesText = ({ children, ...props }) => {
```
joclefrancois, line 44: Can we name it with a full name, i.e., MultipleLinesText please ?

src/components/MaterialUI/NLinesText.js
```diff
@@ -0,0 +1,68 @@
...
+	const classes = useStyles(props)();
+	const textProps = props.textProps;
+
+	if (textProps != null && textProps instanceof TextProps === false) {
+		throw new Error("textProps property is not of type TextProps");
```
joclefrancois, line 49: We have unit tests for that as well, it might be the only missing one actually.

src/components/MaterialUI/NLinesText.test.js
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+import { mount } from "enzyme";
+import NLinesText from "./NLinesText";
+import TextProps from "./textProps";
+import Typography from "@material-ui/core/Typography";
+
+describe("NLineText", () => {
```
joclefrancois, line 7: Ditto

src/components/MaterialUI/NLinesText.test.js
```diff
@@ -0,0 +1,34 @@
...
+			"Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec egestas purus non augue tempor.";
+		const component = (
+			<NLinesText lineCount="2" textProps={nLinesTextProps}>
+				{text}
+			</NLinesText>
+		);
+
+		const mountedComponent = mount(component);
+
+		expect(mountedComponent.exists(".test"), "to be true");
```
joclefrancois, line 32: With ".countains", not sure, but I think you you be able to assert it had been rendered on two lines as expected.

src/components/MaterialUI/textProps.js
```diff
@@ -0,0 +1,22 @@
+import ComponentProps from "./componentProps";
+
+export default class TextProps extends ComponentProps {
+	static propNames = {
+		classes: "classes",
+	};
+
+	static ruleNames = {
+		root: "root",
+		body1: "body1",
```
joclefrancois, line 10: Can we have a better name that "body1" please ?

src/components/MaterialUI/textProps.test.js
```diff
@@ -0,0 +1,12 @@
+import TextProps from "./textProps";
+
+describe("Text Props", () => {
+	it("Sets the style value", () => {
+		const textProps = new TextProps();
+
+		const test = "test";
+		textProps.setStyle(TextProps.ruleNames.body1, test);
+
+		expect(textProps.componentClasses.get("body1"), "to equal", test);
```
joclefrancois, line 10: Add a test also for "root", or change this test...

src/components/MaterialUI/textProps.js
Crispried: @joclefrancois 
We shouldn't. It corresponds to material ui API (yeah they could create a better name for it but, it's body1)

joclefrancois: Oh well, so be it

src/components/MaterialUI/NLinesText.js
```diff
@@ -0,0 +1,68 @@
...
+				content: "'...'",
+				textAlign: "right",
+				bottom: "0",
+				right: "0",
+				width: "25%",
+				display: "block",
+				position: "absolute",
+				height: "calc(1em * 1.2)",
+				background:
+					"linear-gradient(to right, rgba(255, 255, 255, 0), rgba(255, 255, 255, 1) 75 %)",
```
Crispried, line 26: Is that necessary to have linear-gradient in background color here? The color for "dots" should be the same the color for text and used from theme.

src/components/MaterialUI/NLinesText.test.js
Anstsiya: It's not possible to check like this because html renders full text line. And after styles make it in several lines and truncated. But I added test to check if we have proper styles for this on component

PR #46, opened 2020-06-08
Material ui integration

Crispried:


src/components/MaterialUI/Surfaces/ExpansionPanel.js
```diff
@@ -69,9 +71,16 @@ const ExpansionPanel = ({
...
 
 	// Expansion panel props
+	const defaultExpanded = expansionPanelProps?.get(ExpansionPanelProps.propNames.defaultExpanded);
 	const disabled = expansionPanelProps?.get(ExpansionPanelProps.propNames.disabled);
-	const expanded = expansionPanelProps?.get(ExpansionPanelProps.propNames.expanded);
-	const onChange = expansionPanelProps?.get(ExpansionPanelProps.propNames.onChange);
+
+	const [expanded, setExpanded] = React.useState(defaultExpanded);
+	const [internalExpanded, updateViewState] = useViewState(expansionPanelId);
+
```
joclefrancois, line 79: I don't think you need to have both "useState" and "useViewState". The second one should enough to cover your needs.

joclefrancois: Maybe useViewState should have a second optional parameter to set a default when view state does not exist yet, furthermore, your internal state can be initialized with defaultExpanded=true, for exemple, while in Redux, it can be false.

Crispried: @joclefrancois it's enough to make it work however there will be an error in console
image
It's because of this component was made by MUI as controlled and could be updated only by setState

PR #48, opened 2020-06-10
Theme evolution

a-rousseau:
@joclefrancois @TONYFECTEAU @Crispried 
my first PR ! Some changes will be needed by @Anstsiya for her PR concerning font-size not being in the theme.


src/components/MaterialUI/muiThemes.js
```diff
@@ -473,33 +578,84 @@ const setThemeOverrides = theme => ({
...
 		select: {
 			...theme.select,
 			minWidth: theme.spacing(15),
-			borderRadius: 4,
-			"&:focus": {
-				borderRadius: 4,
-				borderColor: "#4fa1f0",
-				boxShadow: "0 0 4px #4fa1f0",
-				outline: "none",
-			},
```
Crispried, line 502: @joclefrancois that global style was overridden by you, please make sure necessary styles weren't removed.

src/components/MaterialUI/muiThemes.js
```diff
@@ -7,9 +7,15 @@ const commonTheme = {
...
-		borderRadius: 4,
+		borderRadius: 4
 	},
 	typography: {
+		h1Size: 24,
+		h2Size: 18,
+		h3Size: 16,
+		h4Size: 14,
+		h5Size: 13,
+		h6Size: 13,
```
Anstsiya, line 18: is it here right for 13? And I don't see 12px that is already used in ExpansionPanel

src/components/MaterialUI/muiThemes.js
```diff
@@ -473,33 +578,84 @@ const setThemeOverrides = theme => ({
...
 		...theme.MuiSelect,
 		select: {
 			...theme.select,
 			minWidth: theme.spacing(15),
-			borderRadius: 4,
-			"&:focus": {
-				borderRadius: 4,
-				borderColor: "#4fa1f0",
-				boxShadow: "0 0 4px #4fa1f0",
-				outline: "none",
```
joclefrancois, line 501: @a-rousseau, you did not want to keep the focus styling, I thought it would be better to keep it, especially about the border radius being 0 when having the focus.

a-rousseau: @joclefrancois the styling was defined in a more global components like MuiInputBase-input, MuiButtonBase-root. so more components will benefit from this

src/components/MaterialUI/muiThemes.js
a-rousseau: @Anstsiya I noticed it was missing when I sent the PR, will add it today. It will be called fieldLabelSize so you could start using it.

src/components/MaterialUI/muiThemes.js
a-rousseau: @Crispried checked this with @joclefrancois so will resolve this

PR #50, opened 2020-06-11
Tooltip

Crispried:


src/components/MaterialUI/DataDisplay/TableCell.js
```diff
@@ -0,0 +1,54 @@
...
+
+  const classes = useStyles({ cellPaddingInPixels });
+
+  const ref = useRef(null);
+
+  const [clientWidth, setWidth] = React.useState(0);
+
+  useEffect(() => {
+    setWidth(ref.current.clientWidth - 2 * cellPaddingInPixels); // left and right padding
+  }, []);
```
joclefrancois, line 25: You need to put some stuff between [], ie, "ref" and "cellPaddingInPixels".

Also, you would need to put a comment about eslint to get rid of the annoying warning about "setWidth".

src/components/MaterialUI/DataDisplay/TableCell.js
```diff
@@ -0,0 +1,54 @@
...
+      classes={{
+        root: classNames(classes.cellRoot),
+      }}
+    >
+      {shouldWrapComponentInTooltip(cell) === true ? <Tooltip value={cell} maxWidth={clientWidth} /> : cell}
+    </TableCellMui>
+  );
+};
+
+export const shouldWrapComponentInTooltip = function (component) {
```
joclefrancois, line 39: I thought you would move and use that method in Tooltip component ?

src/components/MaterialUI/DataDisplay/Tooltip.js
```diff
@@ -0,0 +1,50 @@
...
+  const classes = useStyles({ maxWidth });
+
+  const ref = useRef(null);
+
+  const [isDisplayed, setIsDisplayed] = React.useState(false);
+
+
+  useEffect(() => {
+    setIsDisplayed(ref.current.clientWidth >= maxWidth);
+  }, [maxWidth]);
```
joclefrancois, line 34: Add "ref" between squared brackets and add the eslint comment about "setIsDisplayed"

src/components/MaterialUI/DataDisplay/TableCell.js
joclefrancois: Well, I don't see esling warning but usually, it compains about it, strange. Run "npm run lint" locally to see if you have any. At the same time, could you remove those from MultipleLineText component please ?

src/components/MaterialUI/DataDisplay/TableCell.js
Crispried: Yep we discussed it but, when I think once more I decided that it's not a general behavior. I mean we wan't to wrap cells in Tooltip if they are primitives by default. But, if something was passed in Tooltip component we have to make it tooltip despite on what that component is (if client need to wrap it in tooltip then it should be wrapped). And anyway even primitive should be wrapped in div because of behavior of block element.

src/components/MaterialUI/DataDisplay/Tooltip.js
Crispried: done

PR #54, opened 2020-06-12
Show empty list icon

Anstsiya:
added Placeholder component and added ability to use placeholder in MaterialUI Table

### DIFF
--- a/src/components/MaterialUI/Placeholder.js
+++ b/src/components/MaterialUI/Placeholder.js
@@ -1,0 +1,45 @@
+import React from "react";
+import Icon from "./../Icon";
+import Grid from "@material-ui/core/Grid";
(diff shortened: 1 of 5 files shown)


src/components/MaterialUI/Placeholder.js
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+import Icon from "./../Icon";
+import Grid from "@material-ui/core/Grid";
+import Typography from "@material-ui/core/Typography";
+import { makeStyles } from "@material-ui/core/styles";
+
+const useStyles = makeStyles(theme => ({
+	placeholder: {
+		margin: "50px auto 50px auto",
```
Crispried, line 9: Use theme.spacing for margins/paddings and sizes.

PR #65, opened 2020-07-02
Features/25717 ability to show tooltips for order table columns

Crispried:


src/components/MaterialUI/muiThemes.js
```diff
@@ -676,8 +676,8 @@ const setThemeOverrides = theme => ({
 			border: "none",
 		},
 	},
-	MuiExpansionPanelSummary: {
-		...theme.MuiExpansionPanelSummary,
+	MuiAccordionSummary: {
```
joclefrancois, line 679: Would it hurt to keep both ? We never know...

Crispried: Doesn't make sense. Those blobal classes not exists for Accordion. And they have plans to remove Expansion Panel soon

PR #70, opened 2020-07-07
created Divider component according to code refactoring for bug #38394

Anstsiya:


src/components/MaterialUI/DataDisplay/Divider.js
```diff
@@ -0,0 +1,58 @@
...
+
+	if (dividerProps != null && dividerProps instanceof DividerProps === false) {
+		throw new TypeError("dividerProps property is not of type DividerProps");
+	}
+
+	const orientation = dividerProps?.get(DividerProps.propNames.orientation);
+	const light = dividerProps?.get(DividerProps.propNames.light);
+	const variant = dividerProps?.get(DividerProps.propNames.variant);
+
+	var customStyles = dividerProps?.get(DividerProps.propNames.classes);
```
Crispried, line 31: It's not necessary to have classes in your props. The idea is that instead of accessing classes props you can access styles directly (using getStyle). And to set style use setStyle.

Also replace var with const. So, do smth like

const rootStyle = divierProps?.getStyle(DividerProps.ruleNames.root.

Same for others (like lightStyle, verticalStyle, etc.)

src/components/MaterialUI/DataDisplay/Divider.js
```diff
@@ -0,0 +1,58 @@
...
+					classes.dividerLight,
+					customStyles?.get(DividerProps.ruleNames.light),
+				),
+				vertical: classNames(
+					classes.dividerVertical,
+					customStyles?.get(DividerProps.ruleNames.vertical),
+				),
+				middle: classNames(customStyles?.get(DividerProps.ruleNames.middle)),
+				inset: classNames(customStyles?.get(DividerProps.ruleNames.inset)),
+			}}
```
Crispried, line 53: align with another comment

src/components/MaterialUI/DataDisplay/dividerProps.js
```diff
@@ -0,0 +1,34 @@
+import ComponentProps from "../componentProps";
+
+class DividerProps extends ComponentProps {
+	static propNames = {
+		classes: "classes",
```
Crispried, line 5: You don't need classes here

src/components/MaterialUI/DataDisplay/dividerProps.js
```diff
@@ -0,0 +1,34 @@
...
+		middle: "middle",
+		inset: "inset",
+	};
+
+	constructor() {
+		super();
+		this.componentProps.set(this.constructor.propNames.orientation, null);
+		this.componentProps.set(this.constructor.propNames.light, null);
+		this.componentProps.set(this.constructor.propNames.variant, null);
+		this.componentProps.set(this.constructor.propNames.classes, this.componentClasses);
```
Crispried, line 24: Same, don't need classes

PR #72, opened 2020-07-08
Features/38328 migration of orders table to mui

joclefrancois:
Adapted the Table component to match order list requirements.


src/components/MaterialUI/DataDisplay/tableHelpers.test.js
```diff
@@ -0,0 +1,505 @@
...
+			value: "0",
+		});
+		expect(rows[2].columns[0].title.props, "to equal", {
+			style: "currency",
+			currency: "CAD",
+			value: "0",
+		});
+	});
+
+	it("build table rows as expected with date", () => {
```
TONYFECTEAU, line 328: I notice the data is numeric, but not a date?  Is this a valid test?

src/components/MaterialUI/DataDisplay/tableHelpers.test.js
```diff
@@ -0,0 +1,505 @@
...
+		expect(rows[0].columns.length, "to equal", 1);
+		expect(rows[0].element, "to equal", elements[0]);
+		expect(rows[0].columns[0].cellElement.props, "to equal", { value: 123 });
+		expect(rows[0].columns[0].title.props, "to equal", { value: 123 });
+
+		expect(rows[1].columns[0].cellElement.props, "to equal", { value: 44 });
+		expect(rows[1].columns[0].title.props, "to equal", { value: 44 });
+	});
+
+	it("build table rows as expected with datetime", () => {
```
TONYFECTEAU, line 349: Ditto for datetime.

Crispried:
Resolve lint comments

src/components/MaterialUI/DataDisplay/Table.js
```diff
@@ -1,64 +1,361 @@
...
+		return (
+			<td className={props.className} ref={ref} {...props}>
+				{props.value}
+			</td>
+		);
+	}),
+);
+
+// We will assume for now that cell will never change value so we want to avoid rerendering as much as possible
+export const MemoTableCell = React.memo(TableCell, () => true);
```
Crispried, line 125: Sounds like potential possible "bug" in future :) 
Is that really affects performance so much to have correct compare function right now?

src/components/MaterialUI/DataDisplay/Table.js
```diff
@@ -1,64 +1,361 @@
...
+
+const buildRowCheckbox = (classes, key, selectionHandlers) => {
+	const rowSelected = selectionHandlers.isSelected(key);
+
+	return (
+		<td className={classes.tableCellSelect} selected={rowSelected}>
+			<CheckboxMui
```
Crispried, line 132: Isn't better here to use created wrapper for checkbox? I guess checkbox inside table are the same as others in design.

src/components/MaterialUI/DataDisplay/Table.js
```diff
@@ -1,64 +1,361 @@
...
+				onChange={event => selectionHandlers.selectionHandler(event, key)}
+			/>
+		</td>
+	);
+};
+
+const buildTableCheckbox = (classes, tableSelectionStatus, selectionMethods) => {
+	return (
+		<th key="tableSelector" className={classes.headerCellSelect}>
+			<CheckboxMui
```
Crispried, line 144: Same question

src/components/MaterialUI/DataDisplay/Table.js
```diff
@@ -1,64 +1,361 @@
...
+	selectionMethods,
+) => {
+	const tableCheckbox =
+		selectMode === true
+			? [buildTableCheckbox(classes, tableSelectionStatus, selectionMethods)]
+			: [];
+
+	return tableCheckbox.concat(
+		headers.map((header, index) => (
+			<th key={index} className={classes.headerCell} style={header.cellStyle}>
```
Crispried, line 179: What style property will do here? it's not react right?
You can use several classes inside className. To mix classes you can use classnames package (you can find sample inside ExpansionPanel.js)

src/components/MaterialUI/DataDisplay/Table.js
```diff
@@ -1,64 +1,361 @@
...
+			selected={selectionHandlers.isSelected(row.key)}
+		>
+			{selectMode === true ? buildRowCheckbox(classes, row.key, selectionHandlers) : null}
+			{row.columns.map((cell, cellIndex) => (
+				<MemoTableCell
+					className={classes.tableCell}
+					key={cellIndex}
+					value={cell.cellElement}
+					titleValue={cell.title}
+					style={cell.cellStyle}
```
Crispried, line 207: What style property will do here? it's not react right?
You can use several classes inside className. To mix classes you can use classnames package (you can find sample inside ExpansionPanel.js)

src/components/MaterialUI/DataDisplay/Table.js
```diff
@@ -1,64 +1,361 @@
...
+					rows={props.rows}
+					selectedNumber={props.selectedNumber}
+				/>
+			</TableMui>
+		</div>
+	);
+});
+
+const Table = ({
+	contentTimestamp,
```
Crispried, line 250: is that possible to avoid that parameter? as a user of table I don't really want pass parameters like this. is that possible to get timestamp just inside the table?

src/components/MaterialUI/DataDisplay/Table.js
```diff
@@ -1,64 +1,361 @@
...
+		</TableContainer>
 	);
 };
 
-export default Table;
+function tablePropsAreEqual(prev, next) {
+	return (
+		next.contentTimestamp !== undefined && prev.contentTimestamp === next.contentTimestamp
+	);
+}
```
Crispried, line 359: I understand that it's easy way to detect changes but, in fact it requires a user to pass an extra parameter which is a bit strange (how timestamp related with table data I want to pass as a user).

It would be great to have a HOC which would add that timestamp for a user. Maybe I don't see the whole picture and there is something what allows final user to not think about that property.

src/components/MaterialUI/DataDisplay/TooltippedElements/TooltippedTypography.test.js
```diff
@@ -1,22 +1,34 @@
...
-    const children = "text";
-    const titleValue = "tooltip title";
-    const className = "class";
+	it("should contain the forwarded ref in the child span", () => {
+		const value = "tooltip title";
 
-    const tooltippedTypography = (
-      <TooltippedTypography children={children} className={className} titleValue={titleValue} />
-    );
+		const component = mount(<TooltippedTypography value={value} />);
```
Crispried, line 9: shallow should be enough here btw.

As you maybe noticed I started use shallow as much as possible coz it executes much faster. And since we are having more than 1000 tests now in orc-shared (and probably we gonna have 5 times more to the moment when OMS product will be done) it make sense to start think about saving execution time for tests.

So, if possible to use shallow for test then better use it instead of mount.

src/components/MaterialUI/DataDisplay/tableHelpers.test.js
joclefrancois: Dates are actually numeric values and the purpose of the test is not actually to test the "FormattedDate" or the "FormattedTime" component from react-intl.

src/components/MaterialUI/DataDisplay/tableHelpers.test.js
joclefrancois: Ditto

src/components/MaterialUI/DataDisplay/TooltippedElements/TooltippedTypography.test.js
joclefrancois: I tried with shallow, but test coverage is not 100% in that case, adding a simple test like I did brought it back to 100%. It does not actually hurt, still, I understand your point.

src/components/MaterialUI/DataDisplay/Table.js
joclefrancois: "style" is a react attribute yes, I wanted to include in column definitions (useConfig pattern) a way to provide custom style for a column. I will change my recipies for that.

src/components/MaterialUI/DataDisplay/Table.js
joclefrancois: Ditto

src/components/MaterialUI/DataDisplay/Table.js
joclefrancois: An HOC would add a prop that would be managed internally by the HOC, problem here is the user still must provide it to inform the Table component need to render something new, an HOC would not make any differences here. Even though I myself am not sure to like it that solution, I did not find any other way.

package.json
```diff
@@ -54,5 +55,9 @@
 		"hooks": {
 			"pre-commit": "lint-staged"
 		}
+	},
+	"dependencies": {
+		"@material-ui/core": "^4.11.0",
```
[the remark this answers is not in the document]
joclefrancois: I simply forgot it, will remove.

src/components/MaterialUI/DataDisplay/Table.js
joclefrancois: There is a use case when Filter Bar is opened and it triggers the scroll bar to appear on the table. I would need an resize event listener on the DOM element, which is not supported out-of-the-box by most browsers. What would you think about the package "https://github.com/marcj/css-element-queries" ? Let me know please

src/components/MaterialUI/DataDisplay/Table.js
joclefrancois: This question is related to the other in fact, I will resolve this comment and will come to an agreement on the other one.

package.json
```diff
@@ -54,5 +55,9 @@
...
+		"@welldone-software/why-did-you-render": "^4.2.5"
```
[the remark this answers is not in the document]
joclefrancois: I will take care of that in another story(soon), for now, I will simply remove it.

src/components/MaterialUI/DataDisplay/Table.js
joclefrancois: I thought about it, don't worry, problem is I cannot compare "value" as it may be component, which could be different everytime even though it might not be in fact. I would need primitive type to ensure it is not rerendered everytime. Will do some more tests.

src/components/MaterialUI/DataDisplay/Table.js
joclefrancois: The only reason why I had created a wrapper for Checkbox in the first place was to encapsulate "FormControlLabel" in order to avoid that hassle everytime we need it. There would no adding value to use our wrapper as we don't need the label for this case.

src/components/MaterialUI/DataDisplay/Table.js
joclefrancois: Same answer.

src/components/MaterialUI/DataDisplay/Table.js
Crispried: Yeah understand the problem with components.. Maybe it won't hurt to much if cell components will be rerenderred. Also, what I'm suggesting as "naive" compare scenario for components we can implement is to try check if all children and value properties inside that component are the same (most of the time values are inside those 2 properties). For that I would suggest to implement helper function which will retrieve values for all the "children" and "value" properties for passed component.

src/components/MaterialUI/DataDisplay/Table.js
Crispried: Well, if in this case we need to change only styles based on resize event then I would say it would be better to have packages like this

src/components/MaterialUI/DataDisplay/Table.js
joclefrancois: Using TableCell directly is fine enough after some exhaustive testing. I will remove MemoTableCell as maybe it is some legacy code of an earlier design or previous testing I had performed.

src/components/MaterialUI/DataDisplay/Table.js
joclefrancois: Let me try it first, if it works, I will add it in orc-script as well as why-did-you-rerender while I am at it.

PR #76, opened 2020-07-15
1. badge was added 2. segment label now could be component as well

Crispried:


src/components/Routing/SegmentPage.js
```diff
@@ -41,16 +41,16 @@ export const Item = styled(FilteredLink)`
 	color: ${getThemeProp(["colors", "text"], "#333333")};
 
 	${ifFlag(
```
joclefrancois, line 43: Indentation looks weird here, maybe it's only on github.

PR #77, opened 2020-07-16
moved placeholder

Anstsiya:


src/components/MaterialUI/DataDisplay/Table.js
```diff
@@ -85,6 +85,12 @@ export const useStyles = makeStyles(theme => ({
 	rowSelectCheckbox: {
 		padding: 0,
 	},
+	placeholder: {
+		position: "fixed",
```
Crispried, line 89: I would suggest that absolute position is more appropriate here.

PR #79, opened 2020-07-17
Add Radio component

benaube:
Add Radio component
Add ratio tool to ToolBar

### DIFF
--- a/src/components/MaterialUI/Inputs/Radio.js
+++ b/src/components/MaterialUI/Inputs/Radio.js
@@ -5,7 +5,6 @@ import RadioMui from "@material-ui/core/Radio";
 import FormControlLabel from "@material-ui/core/FormControlLabel";
 import FormControl from "@material-ui/core/FormControl";
 import FormLabel from "@material-ui/core/FormLabel";
-import { IntlContext } from "react-intl";
 import { makeStyles } from "@material-ui/core/styles";
 import RadioProps from "./RadioProps.js";
 
@@ -38,8 +37,6 @@ const useStyles = makeStyles(theme => ({
 	},
 }));
 
-const maybeTranslate = (formatMessage, message) => (message?.id ? formatMessage(message) : message);
-
 const extractAndValidateProps = radioProps => {
 	if (radioProps != null && radioProps instanceof RadioProps === false) {
 		throw new TypeError("radioProps property is not of type RadioProps");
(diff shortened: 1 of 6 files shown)


src/components/MaterialUI/Inputs/Radio.js
```diff
@@ -0,0 +1,113 @@
...
+			position: "absolute",
+			borderRadius: "50%",
+			top: "50%",
+			left: "50%",
+			transform: "translate(-50%, -50%)",
+		},
+	},
+}));
+
+const maybeTranslate = (formatMessage, message) => (message?.id ? formatMessage(message) : message);
```
joclefrancois, line 41: Do we have such pattern somewhere else? Should we not assume that all labels will already be localized like in the Checkbox or the Table components ? I would say it is the responsibility of the component user to make sure the label is set properly.

benaube: In the OMS Toolbar, nothing was translated. Both the button and input are translated in orc-shared.
Input.js does the MaybeTranslate.
IconButton uses Text.js which does the translation.

src/components/MaterialUI/Inputs/Radio.js
```diff
@@ -0,0 +1,113 @@
...
+				value={value}
+				onChange={handleChange}
+			>
+				{radios.map((radio, index) => {
+					const handleClick = radio.clickEvent ? event => radio.clickEvent(event.target.value) : null;
+
+					return (
+						<FormControlLabel
+							key={index}
+							name={radio.value}
```
joclefrancois, line 94: I am guessing radio.value should be unique in the collection of radios. Did you check that out? If yes, I would also use it as a key with a prefix, something like "radiobutton_" + radio.value. It is usually not recommended to use index as a key, I agree that sometimes it's hard to use something else but whenever we can, we should use something else.

src/components/MaterialUI/Inputs/Radio.js
joclefrancois: I would say that this is not the approach we take for Material-UI wrapper components we create. We want to eventually deprecate Gert's components you are referring to. I understand you point though but we should keep the same recipe for all of our wrappers. I am not saying the current one is the best one, I am just saying we should keep consistency throughout our code when related to Material-UI. Even Gert's toolbar might be replaced eventually.

benaube: Alright, so you'd rather Radio doesn't translate and i do the translation in OMS useConfig where the tools are defined?

joclefrancois: I would say yes, especially that the label can also be a component or something else. I wouldn't give the responsibility of translation to our components.

joclefrancois:
Look at the coverage using the following link, if not a big deal, I would try to add or change a unit test to have coverage at 100%.

https://travis-ci.com/github/Orckestra/orc-shared/jobs/363386869

PR #80, opened 2020-07-21
Features/38725 show execution message icon in summary for orders flagged

Crispried:


src/content/icons/dropdawn-chevron-down.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
joclefrancois, line 1: Did you mean "dropdown" ? It is not the Dawn of the Dead as far as I know 😉

src/components/MaterialUI/Inputs/Select.js
```diff
@@ -57,7 +57,7 @@ const MenuProps = {
 };
 
 const ChevronDown = props => {
-	return <Icon id="chevron-down" {...props} />;
+	return <Icon id="dropdawn-chevron-down" {...props} />;
```
joclefrancois, line 60: To be renamed along with file.

src/components/MaterialUI/hocs/withDeferredPopper.js
```diff
@@ -0,0 +1,127 @@
...
+    },
+  },
+  arrow: {
+    position: "absolute",
+    border: "1px solid",
+    backgroundColor: theme.palette.background.paper,
+  }
+}));
+
+export const Arrow = () => {
```
joclefrancois, line 59: You can also set classes as Props of Arrow if you want, up to you. This is what I did for the Table to avoid to use the hook at every level.

joclefrancois:
One warning to be cleaned.

src/content/icons/dropdawn-chevron-down.svg
Crispried: Not me, that was Alain who gave names for those files I've just imported them from zeroheight. I will correct

src/components/MaterialUI/hocs/withDeferredPopper.js
```diff
@@ -0,0 +1,127 @@
...
+  };
+
+  const clickAwayHandler = function () {
+    setPopperState({
+      isDisplayed: false,
+      anchorElement: null
+    })
+  }
+
+  if (popperValue == null) return <Comp {...props} />;
```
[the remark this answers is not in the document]
Crispried: In fact before mounting (what happens after return) it doesn't hurt performance to execute all those lines before. From the code readability probably it would be better to handle that as early as possible.

src/components/MaterialUI/hocs/withDeferredPopper.js
Crispried: agree

src/components/MaterialUI/hocs/withDeferredPopper.js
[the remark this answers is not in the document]
joclefrancois: Oh, I was not saying it would hurt performance, it was only about sequence of operations as I personal preference of mine.

PR #81, opened 2020-07-27
added reusable address component

Anstsiya:


src/components/MaterialUI/DataDisplay/Address.js
```diff
@@ -0,0 +1,46 @@
...
+	line2,
+	city,
+	province,
+	postalCode,
+	country,
+	email,
+	phone,
+	textProps,
+}) => {
+	const classes = useStyles();
```
joclefrancois, line 23: Not use anywhere, you can remove it. (lint gives a warning about it)

src/components/MaterialUI/DataDisplay/Address.js
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import Grid from "@material-ui/core/Grid";
+import { makeStyles } from "@material-ui/core/styles";
+import { TextProps } from "../textProps";
+import TooltippedTypography from "./TooltippedElements/TooltippedTypography";
+import _ from "lodash";
+
+const useStyles = makeStyles(theme => ({}));
```
joclefrancois, line 8: If not used, remove it.

src/components/MaterialUI/DataDisplay/Address.js
```diff
@@ -0,0 +1,46 @@
...
+
+const Address = ({
+	name,
+	contactPerson,
+	line1,
+	line2,
+	city,
+	province,
```
joclefrancois, line 16: "province" is maybe too much Canada-Oriented as it may have different names in other countries. In our database, we call that "regionName", it would be more a generic term.

src/components/MaterialUI/DataDisplay/Address.js
```diff
@@ -0,0 +1,46 @@
...
+	phone,
+	textProps,
+}) => {
+	const classes = useStyles();
+
+	if (textProps != null && textProps instanceof TextProps === false) {
+		throw new TypeError("textProps property is not of type TextProps");
+	}
+
+	const cityProvincePostalCode = `${city}, ${province}, ${postalCode}`;
```
joclefrancois, line 29: Ditto

PR #87, opened 2020-08-03
Created a search control for orders and any other entities in the future

joclefrancois:
Created a search control for orders and any other entities in the future.

Story: #25140


package.json
```diff
@@ -37,6 +37,7 @@
 	},
 	"homepage": "https://github.com/Orckestra/orc-shared#readme",
 	"devDependencies": {
+    "jest": "^26.1.0",
```
Crispried, line 40: Remove that dependency

Crispried:
Move SearchControl to Input/PredefinedElements

src/components/MaterialUI/Inputs/SearchControl.js
```diff
@@ -0,0 +1,175 @@
...
+import Icon from "./../DataDisplay/Icon";
+import IconButton from "@material-ui/core/IconButton";
+import Select from "./Select";
+import Input from "@material-ui/core/Input";
+
+export const useStyles = makeStyles(theme => ({
+	clearButton: {
+		fontSize: theme.spacing(2.2),
+		padding: theme.spacing(0.3),
+		marginLeft: "0 !important",
```
Crispried, line 13: Will it work without important? It should

src/components/MaterialUI/Inputs/SearchControl.js
```diff
@@ -0,0 +1,175 @@
...
+		},
+		"&:focus": {
+			boxShadow: "none",
+			borderRadius: 0,
+		},
+	},
+	searchButton: {
+		fontSize: theme.spacing(2.2),
+		padding: `${theme.spacing(0.3)} ${theme.spacing(1)}`,
+		marginLeft: "0 !important",
```
Crispried, line 29: ditto

src/components/MaterialUI/Inputs/Select.js
```diff
@@ -93,26 +78,28 @@ const Select = ({ options, selectProps }) => {
 
 	const update = selectProps?.get(SelectProps.propNames.update);
 	const value = selectProps?.get(SelectProps.propNames.value);
-	const numericSort = selectProps?.get(SelectProps.propNames.numericSort) || false;
-	const showAllValue = selectProps?.get(SelectProps.propNames.showAllValue) || "#All#";
-	const showAllLabel = selectProps?.get(SelectProps.propNames.showAllLabel) || "Show All";
+	const sortType = selectProps?.get(SelectProps.propNames.sortType) || sortTypeEnum.none;
+	const showAllValue = selectProps?.get(SelectProps.propNames.showAllValue);
+	const showAllLabel = selectProps?.get(SelectProps.propNames.showAllLabel);
+	const classesProps = selectProps?.get(SelectProps.propNames.classes);
```
Crispried, line 84: Classes shouldn't be defined through props. The rule names should be defined through componentClasses and then use getStyle and setStyle. So, for this case ruleNames could be menuPropsPaper and root. Check dividerProps for sample

src/components/MaterialUI/Inputs/SelectProps.js
```diff
@@ -1,21 +1,29 @@
...
 class SelectProps extends ComponentProps {
 	static propNames = {
 		update: "update",
 		value: "value",
-		numericSort: "numericSort",
+		sortType: "sortType",
 		showAllValue: "showAllValue",
-		showAllLabel: "showAllLabel"
+		showAllLabel: "showAllLabel",
+		classes: "classes",
```
Crispried, line 16: Remove classes from props

src/components/MaterialUI/Inputs/SelectProps.js
```diff
@@ -1,21 +1,29 @@
...
 
 	constructor() {
 		super();
 		this.componentProps.set(this.constructor.propNames.update, null);
 		this.componentProps.set(this.constructor.propNames.value, null);
-		this.componentProps.set(this.constructor.propNames.numericSort, null);
+		this.componentProps.set(this.constructor.propNames.sortType, null);
 		this.componentProps.set(this.constructor.propNames.showAllValue, null);
 		this.componentProps.set(this.constructor.propNames.showAllLabel, null);
+		this.componentProps.set(this.constructor.propNames.classes, null);
```
Crispried, line 26: ditto

src/components/MaterialUI/Inputs/SearchControl.js
joclefrancois: The style from "MuiInputBase-root" takes precedence if I don't put important. Any suggestions ? I would rather not changing it in muiThems.js of course.

src/components/MaterialUI/Inputs/SearchControl.js
joclefrancois: ditto

src/components/MaterialUI/Inputs/SearchControl.js
Crispried: Hm, ok.. that's strange coz we have many places were we are overriding global classes by this way without important. but, if it doesn't work without it left it as it is. I don't think it make sense to spend time on it.

PR #94, opened 2020-08-24
Update DatePicker styling - related to task 38293 of user Story 25145

a-rousseau:


a-rousseau:
Having this fail where Intlprovider is TypeError: intl_1.createIntlCache is not a function

a-rousseau:
Going to start over

PR #163, opened 2020-12-22
generic approach to retrieve entityId from url was implemented

Crispried:


InnaBoitsun (approved):
Tested in Settings App

PR #170, opened 2021-01-04
Add the checkbox feature

khanvong-ork:
Add the X red icon when readonly and value is false and use the  default checkmark when value is true
image


Anstsiya:
Seems you did no changes for unit tests. Update them as well please.

src/components/MaterialUI/Inputs/CheckboxProps.js
```diff
@@ -5,13 +5,17 @@ class CheckboxProps extends ComponentProps {
 		update: "update",
 		value: "value",
 		label: "label",
+		readOnly: "readOnly",
+		checked: "checked",
```
joclefrancois, line 9: Why did you add "checked" as "value" is already there ?

Anstsiya: Seems that checked and value it's different props. Value is for the value checkbox represents, but checked for Boolean value of that value

joclefrancois: @Anstsiya, yes, but for the Wrapper, value was already fulfilling that purposes, we don't need that duplication.

Anstsiya: @joclefrancois  but don't we need to know what value was changed and put it into a handler to have it in edit mode?

joclefrancois: Furthermore, it is a breaking changes for any application using it already.

src/components/MaterialUI/Inputs/Checkbox.js
```diff
@@ -44,35 +45,98 @@ const useStyles = makeStyles(theme => ({
...
 const Checkbox = ({ checkboxProps }) => {
 	if (isCheckboxProps(checkboxProps) === false) {
 		throw new TypeError("checkboxProps property is not of type CheckboxProps");
 	}
 
+	const checked = checkboxProps?.get(CheckboxProps.propNames.checked) || false;
 	const update = checkboxProps?.get(CheckboxProps.propNames.update);
 	const value = checkboxProps?.get(CheckboxProps.propNames.value);
 	const label = checkboxProps?.get(CheckboxProps.propNames.label);
+	const readOnly = checkboxProps?.get(CheckboxProps.propNames.readOnly) || false;
```
joclefrancois, line 111: You need to change the file "CheckboxProps" as well and its unit tests.

joclefrancois (changes requested):
See me other comments

src/components/MaterialUI/Inputs/CheckboxProps.js
khanvong-ork: Removed checked so we don't add possible breaking changes. We should have further discussion on why we need the checked attribute.

src/components/MaterialUI/Inputs/Checkbox.js
```diff
@@ -1,9 +1,10 @@
 import React from "react";
-import clsx from 'clsx';
+import clsx from "clsx";
```
joclefrancois, line 2: You don't use it

joclefrancois (approved):

src/components/MaterialUI/Inputs/Checkbox.js
khanvong-ork: Removed the referenced

Crispried (approved):

PR #174, opened 2021-01-12
edit mode fix

oleksandr-honcharuk:
[AB#43106](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/43106)


Crispried:
I don't really understand what does it changes coz in current code sections should be undefned in such case and then condition sections != null won't be passed. I'm expecting to provide steps to reproduce coz I would like to see the sections value in that case.

oleksandr-honcharuk:
Steps to reproduce
On some module, for example Countries, open country details and start EDIT
add some changes - do not close TAB
go to other module, for example Templates and try to open some templates  in  details TAB

Crispried (approved):

PR #178, opened 2021-01-20
Highlight sections with errors

InnaBoitsun:
[AB#43266](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/43266)

- it is possibe to save error for model fields for  specific section so error can be displayed near the field
- when field change value - the error is cleared
- if section has any error, then it is higlighed like this (to be confirmed with designer)

image
section-with-error-edit

### DIFF
--- a/src/components/Routing/SegmentPage.js
+++ b/src/components/Routing/SegmentPage.js
@@ -30,12 +30,6 @@ const useStyles = makeStyles(theme => ({
 		fontSize: theme.typography.fontSize,
 		maxWidth: theme.spacing(15),
 	},
-	errorLabel: {
-		color: theme.palette.error.main,
-		fontWeight: theme.typography.fontWeightSemiBold,
-		fontSize: theme.typography.fontSize,
-		maxWidth: theme.spacing(15),
-	},
 	labelComponent: {
 		margin: `0 ${theme.spacing(1)}`,
 	},
--- a/src/components/Routing/SegmentPage.js
+++ b/src/components/Routing/SegmentPage.js
@@ -89,7 +89,9 @@ const SegmentPage = ({ path, component: View, segments, location, match, moduleP
 		subpages = [];
 	const entityIdKey = Object.keys(match.params).find(p => p !== "scope");
 	let entityId = match.params[entityIdKey];
-	if (!entityId) entityId = tryGetNewEntityIdKey(baseHref);
+	if (!entityId) {
+		entityId = tryGetNewEntityIdKey(baseHref);
+	}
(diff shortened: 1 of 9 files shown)


src/components/Routing/SegmentPage.js
```diff
@@ -150,7 +157,14 @@ const SegmentPage = ({ path, component: View, segments, location, match, moduleP
 
 								const finalLabel = (
 									<Grid container justify="space-between">
-										<Grid item className={classes.label}>
+										<Grid
+											item
+											className={
+												sectionsWithErrors.includes(segpath.replace("/", "")) === true
```
oleksandr-honcharuk, line 163: why need compare with true?

Crispried:
In general I'm good with the approach to store errors for each edit model on the level where we are storing the value for specific field. However in current implementation the logic is that we are going to pass error message based on some field validation which will be determined on client side. And I'm just curious - can we predict all the possible error messages for different field types based on something (like field name, etc.). I'm asking that coz it seems for me on the first point of view that it's possible to generalize errors and as result avoid the case when we will create and pass similar errors for each field in edit model accross all the applications.

Crispried (approved):

src/utils/propertyHelper.test.js
```diff
@@ -109,3 +109,39 @@ describe("isObjectContainsPropertyWithValue", () => {
...
+describe("isObjectContainsPropertyWithAnyValue", () => {
+	it("Returns true if object contains specified property with any value", () => {
+		const obj = {
+			prop1: {
+				a: {
+					m: "Hello",
+				},
+			},
+			prop2: {
+				m: "Hello Bobbina",
```
TONYFECTEAU, line 122: Who is Bobbina?  ;)

TONYFECTEAU (approved):

src/utils/propertyHelper.test.js
Crispried: Female adaptation of Bob based on slavic languages I guess :)

InnaBoitsun:
> In general I'm good with the approach to store errors for each edit model on the level where we are storing the value for specific field. However in current implementation the logic is that we are going to pass error message based on some field validation which will be determined on client side. And I'm just curious - can we predict all the possible error messages for different field types based on something (like field name, etc.). I'm asking that coz it seems for me on the first point of view that it's possible to generalize errors and as result avoid the case when we will create and pass similar errors for each field in edit model accross all the applications.

I was thinking about some common FieldRequired error - we can think  about it later

PR #179, opened 2021-01-20
Added drop down menu and predefined action menu

oleksandr-honcharuk:
[AB#38804](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/38804)

image


src/components/MaterialUI/Navigation/DropDownMenu.js
```diff
@@ -0,0 +1,81 @@
+import React, { useState, isValidElement, cloneElement } from "react";
+import { withStyles } from "@material-ui/core/styles";
+import Button from "@material-ui/core/Button";
+import Menu from "@material-ui/core/Menu";
+import MenuItem from "@material-ui/core/MenuItem";
+import Text from "../../Text";
+
+const CustomMenuItem = withStyles(theme => ({
```
Crispried, line 8: Don't use withStyles. use always makeStyles

src/components/MaterialUI/Navigation/DropDownMenu.js
```diff
@@ -0,0 +1,81 @@
...
+		borderRadius: 0,
+		"&:hover, &:focus": {
+			borderRadius: 0,
+			boxShadow: "none",
+			backgroundColor: theme.palette.primary.light,
+		},
+	},
+}))(MenuItem);
+
+const CustomMenu = withStyles(theme => ({
```
Crispried, line 22: Don't use withStyles. use always makeStyles

src/components/MaterialUI/Navigation/PredefinedElements/ActionMenu.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { withStyles } from "@material-ui/core/styles";
+import IconButton from "@material-ui/core/IconButton";
+import DropDownMenu from "../DropDownMenu";
+import Icon from "../../DataDisplay/Icon";
+
+const LightIconButton = withStyles(theme => ({
```
Crispried, line 7: Don't use withStyles. use always makeStyles

src/components/MaterialUI/Navigation/DropDownMenu.js
```diff
@@ -0,0 +1,81 @@
...
+			transformOrigin: {
+				vertical: "top",
+				horizontal: "right",
+			},
+		}}
+		{...props}
+	/>
+));
+
+const DropDownMenu = ({ payload, menuItems, children, autoFocus = true, ...props }) => {
```
Crispried, line 46: Please create props class for it and pass all the secondary props using it. And try to avoid property spreading for components props for all the properties which are predictable.

src/components/MaterialUI/Navigation/DropDownMenu.test.js
```diff
@@ -0,0 +1,65 @@
...
+		expect(button.length, "to equal", 1);
+
+		button.simulate("click");
+
+		const items = component.find("li");
+		expect(items.length, "to equal", 2);
+
+		items.at(0).simulate("click");
+
+		expect(menuItems[0].action.mock.calls.length, "to equal", 1);
```
Crispried, line 63: it doesn't seems to be a good way to test action. you can find many example in other compoentns for example Switch.test.js which uses unexpected and "to have calls satisfying" function. For handlers which doesn't have any args they should receive you can use just "was called" function. Example is in Modal.test.js.

Anyway the idea is to attach spied callback, then invoke click event (using simulate or invoke methods from enzyme) and then check for that spy to be called.

src/components/MaterialUI/Navigation/DropDownMenu.js
```diff
@@ -0,0 +1,89 @@
...
+		},
+	},
+	menu: {
+		border: `${theme.spacing(0.1)} solid ${theme.palette.grey.borders}`,
+		boxShadow: `${theme.spacing(0, 0.5, 0.7, 0)} ${theme.palette.grey.borders}`,
+		borderRadius: theme.spacing(0.3),
+	},
+}));
+
+const DropDownMenu = ({ payload, menuItems, children, dropDownMenuProps = new DropDownMenuProps() }) => {
```
Crispried, line 28: I have a concern about creating new prop instance for default value of props. I just don't understand what for.. And another thing you missed is actually to check the passed props parameter if it's correct. You've created isDropDownMenuProps  but, didn't use here for some reason. Again the pattern which is used to check it could be found in other wrappers. Also don't forget to add extra unit test for it (samples of such unit tests presents for other wrappers as well).

Crispried (approved):

PR #187, opened 2021-01-27
Extend placeholder for tables

navidehsa:
This is related to https://dev.azure.com/orckestra001/OrckestraCommerce/_workitems/edit/43362/

### DIFF
--- a/src/components/MaterialUI/DataDisplay/PredefinedElements/Placeholder.js
+++ b/src/components/MaterialUI/DataDisplay/PredefinedElements/Placeholder.js
@@ -7,7 +7,7 @@ import Skeleton from "@material-ui/lab/Skeleton";
 
 const useStyles = makeStyles(theme => ({
 	placeholder: {
-		margin: `${theme.spacing(5)} auto`,
+		margin: `${theme.spacing(10)} auto`,
 	},
 	placeholderImage: {
 		fontSize: theme.spacing(14),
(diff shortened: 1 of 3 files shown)


src/components/MaterialUI/DataDisplay/PredefinedElements/Placeholder.js
```diff
@@ -20,11 +21,55 @@ const useStyles = makeStyles(theme => ({
...
 
-const Placeholder = ({ icon, title, subtitle }) => {
+const Placeholder = ({ icon, title, subtitle, cellList = [] }) => {
 	const classes = useStyles();
-	return (
+	return cellList.length ? (
+		<div className={classes.placeholderRow}>
+			{cellList.map((value, index) => {
+				switch (value) {
+					case "status":
```
InnaBoitsun, line 57: Let's name it not status but "chip" as generaly we use here MUI Chip component. Thanks

src/components/MaterialUI/DataDisplay/PredefinedElements/Placeholder.js
```diff
@@ -20,11 +21,55 @@ const useStyles = makeStyles(theme => ({
...
+		animation: false,
+		borderRadius: "15px",
+	},
 }));
 
-const Placeholder = ({ icon, title, subtitle }) => {
+const Placeholder = ({ icon, title, subtitle, cellList = [] }) => {
 	const classes = useStyles();
-	return (
+	return cellList.length ? (
```
InnaBoitsun, line 53: From this logic I see that when we have cellList specified, then Icon and Title will not be rendered. We should be able to render all together  when specifiend like this:
image

InnaBoitsun (approved):

PR #189, opened 2021-01-27
Fix default value check

oleksandr-honcharuk:
[AB#42212](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/42212)

### DIFF
--- a/src/components/MaterialUI/Inputs/Radio.js
+++ b/src/components/MaterialUI/Inputs/Radio.js
@@ -50,9 +50,9 @@ const extractAndValidateProps = radioProps => {
 	const defaultVal = radioProps.get(RadioProps.propNames.defaultVal);
 	const value = radioProps.get(RadioProps.propNames.value) ?? defaultVal;
 	const update = radioProps.get(RadioProps.propNames.update);
-	const row = radioProps.get(RadioProps.propNames.row) || true;
+	const row = radioProps.get(RadioProps.propNames.row) ?? true;
 	const radios = radioProps.get(RadioProps.propNames.radios);
-	const disabled = radioProps.get(RadioProps.propNames.disabled) || false;
+	const disabled = radioProps.get(RadioProps.propNames.disabled) ?? false;
 
 	if (!radios || radios.length < 2) {
 		throw new Error("Radio component must have at least two options");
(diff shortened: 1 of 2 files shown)


src/components/MaterialUI/Inputs/Radio.js
```diff
@@ -50,9 +50,9 @@ const extractAndValidateProps = radioProps => {
 	const defaultVal = radioProps.get(RadioProps.propNames.defaultVal);
 	const value = radioProps.get(RadioProps.propNames.value) ?? defaultVal;
 	const update = radioProps.get(RadioProps.propNames.update);
-	const row = radioProps.get(RadioProps.propNames.row) || true;
+	const row = radioProps.get(RadioProps.propNames.row) ?? true;
```
TONYFECTEAU, line 53: The || could be used, as any false-like value will then be set to true.

src/components/MaterialUI/Inputs/Switch.js
```diff
@@ -45,13 +45,13 @@ export const useStyles = makeStyles(theme => ({
 		"&:before": {
 			content: props => `"${props.formattedOnCaption}"`,
 			left: theme.spacing(0.9),
-			right: props => (props.readOnly ? theme.spacing(0.9) : 0),
+			right: props => props.readOnly && theme.spacing(0.9),
```
TONYFECTEAU, line 48: Hm... this may need to be checked... the prior conditional operator was setting it to theme.spacing(0.9) or to 0.  Now, you are setting it to true or theme.spacing(0.9)

src/components/MaterialUI/Inputs/Radio.js
InnaBoitsun: @TONYFECTEAU row property is **boolean**, if true render group as row, if false render verticaly. 
When you assign row=false - you will have anyway true.
So correct is to use ?? operator, so if it is undefined or null, then it is true.

src/components/MaterialUI/Inputs/Switch.js
oleksandr-honcharuk: false or  theme.spacing(0.9)
this readOnly props change broke current view

InnaBoitsun: Current issue what Oleksand fix here image

InnaBoitsun (approved):

PR #198, opened 2021-02-08
Fix spaces when 2 placeholders

khanvong-ork:


src/components/MaterialUI/DataDisplay/PredefinedElements/Placeholder.js
```diff
@@ -78,11 +78,13 @@ const Placeholder = ({ icon, title, subtitle, cellList = [] }) => {
 						})}
 					</div>
 				) : null}
-				<Grid container direction="column" alignItems="center" className={classes.placeholder}>
-					{icon ? <Icon className={classes.placeholderImage} id={icon} /> : null}
-					{title ? <Typography className={classes.placeholderTitle}>{title}</Typography> : null}
-					{subtitle ? <Typography className={classes.placeholderSubtitle}>{subtitle}</Typography> : null}
-				</Grid>
+				{(icon || title || subtitle) && (
```
TONYFECTEAU, line 81: Hm... there should be a test for this change?

TONYFECTEAU (dismissed):

src/components/MaterialUI/DataDisplay/PredefinedElements/Placeholder.js
khanvong-ork: We don't need to render the Grid when we have only the cellList. A test existed that was validating that we had the Skeleton.

TONYFECTEAU (approved):

PR #201, opened 2021-02-10
Features/41112 edit location attributes

Crispried:


src/components/MaterialUI/DataDisplay/PredefinedElements/InformationItem.js
```diff
@@ -26,10 +25,14 @@ const useStyles = makeStyles(theme => ({
...
 	},
+	container: {
+		width: "100%",
+		marginTop: props => theme.spacing(props.mt),
+	},
 }));
 
-const InformationItem = ({ label, children, required, error, showNotAvailable = false }) => {
-	const classes = useStyles({ required, error });
+const InformationItem = ({ label, children, required, error, showNotAvailable = false, mt = 0 }) => {
```
TONYFECTEAU, line 34: Curious, what does mt stand for?  Margin Top?  Maybe the prop name can be marginTop?

src/selectors/authentication.js
```diff
@@ -1,5 +1,27 @@
 import { createSelector } from "reselect";
+import { platformRoles } from "./../constants";
```
TONYFECTEAU, line 2: Is there a reason for the ./?  Doesn't .. target current directory?

TONYFECTEAU (approved):

src/selectors/authentication.js
```diff
@@ -1,5 +1,27 @@
...
+export const selectGroupRolesClaims = roleGroup =>
+	createSelector(selectRolesClaims, rolesClaims => rolesClaims.get(roleGroup) || null);
+
+export const hasEditorPermissions = roleGroup =>
+	createSelector(selectGroupRolesClaims(roleGroup), appRolesClaims => {
+		if (appRolesClaims != null) {
+			return appRolesClaims.getIn(["*", platformRoles.Editor]);
+		}
+		return false;
+	});
```
InnaBoitsun, line 19: We need to consider the situation that User can have roles just for specific scope, so we also need to pass current scope here: 
"*"  - it means that Scope=Global - user has rigths for all Scopes
But user can have roles just for some scope:
image

Probaly if user does not have Editor permissions for * (Global)  AND for Current Scope - then return FALSE

src/selectors/authentication.js
```diff
@@ -1,5 +1,27 @@
...
+
+export const hasAdministratorPermissions = roleGroup =>
+	createSelector(selectGroupRolesClaims(roleGroup), appRolesClaims => {
+		if (appRolesClaims != null) {
+			return appRolesClaims.getIn(["*", platformRoles.Administrator]);
```
InnaBoitsun, line 24: same here, "*" = GLobal Scope - rights for all scopes

InnaBoitsun (changes requested):

src/translations/en-US.json
```diff
@@ -1,36 +1,38 @@
...
+  "orc-shared.save": "Save",
+  "orc-shared.scopeFilterPlaceholder": "Type a scope name",
+  "orc-shared.showFewerLanguages": "Show fewer languages",
+  "orc-shared.showMoreLanguages": "Show more languages",
+  "orc-shared.signOut": "Sign out",
+  "orc-shared.sortOrder": "Sort Order",
+  "orc-shared.status": "Status",
+  "orc-shared.unsavedChanges": "This entity has unsaved changes. All changes will be lost.\n\nAre you sure you want to close it?",
+  "orc-shared.valueName": "Value Name",
 	"orc-shared.next": "Next",
```
oleksandr-honcharuk, line 36: use same indent for all strings

Crispried: It's prettify configs which are defined in orc-scripts. I'm not doing any manual formatting (my ide config is set to use prettify formatting automatically) and if you would open that file you will see the indents are the same for all the file

src/selectors/authentication.js
Crispried: Just my habbit. Some time ago I found that there is no one "correct" rule in conventions to specify relative paths. Some people like to write relative path implicitly starting from "current directory" others just goes to the top without that. 

Since we have lot of that accross our projects I would prefer to leave things as they are. And if we would decide our internal "rules" about that, then we need to replace it in many places.

src/selectors/authentication.js
Crispried: done

src/selectors/authentication.js
Crispried: done

InnaBoitsun (approved):

PR #203, opened 2021-02-12
Feature/43610 transfer list

oleksandr-honcharuk:


src/components/MaterialUI/DataDisplay/TransferList.js
```diff
@@ -0,0 +1,167 @@
...
+	title: {
+		color: theme.palette.text.hint,
+		fontSize: theme.spacing(1.2),
+		marginBottom: theme.spacing(0.6),
+	},
+	paper: {
+		overflow: "auto",
+		width: theme.spacing(32.6),
+		height: theme.spacing(48),
+		border: `${theme.spacing(0.1)} solid ${theme.palette.primary.light}`,
```
InnaBoitsun, line 28: left papaer should have border: `${theme.spacing(0.1)} solid ${theme.palette.grey.borders}`,
right papaer should have border: `${theme.spacing(0.1)} solid ${theme.palette.primary.light}`

src/components/MaterialUI/DataDisplay/TransferList.js
```diff
@@ -0,0 +1,167 @@
...
+		<List component="div" role="list">
+			{items.map(({ id, title, disabled }) => (
+				<ListItem
+					className={classNames(classes.item, checked.includes(id) && classes.selectedItem)}
+					key={id}
+					role="listitem"
+					onClick={!disabled ? () => toggle(id) : null}
+					disabled={disabled}
+				>
+					<ListItemText id={`transfer-list-item-${id}-label`} primary={title} />
```
InnaBoitsun, line 95: please also add ability to render secondary content:
 <ListItemText id={`transfer-list-item-${id}-label`} primary={title} **secondary={content}** />

src/components/MaterialUI/DataDisplay/TransferList.js
```diff
@@ -0,0 +1,167 @@
...
+	leftListData,
+	onChange,
+	addButtonTitle,
+	deleteButtonTitle,
+	customLeftComponent,
+}) => {
+	const classes = useStyles();
+	const [checked, setChecked] = useState([]);
+	const handleToggle = useCallback(value => setChecked(state => formatSelectedItems(state, value)), [setChecked]);
+	const onTopButtonClick = () =>
```
InnaBoitsun, line 113: please rename to onAddButtonClick

src/components/MaterialUI/DataDisplay/TransferList.js
```diff
@@ -0,0 +1,167 @@
...
+		onChange(formatState({ right: rightListData.data, left: leftListData.data }, checked, "left", "right"));
+	const onBottomButtonClick = () =>
```
InnaBoitsun, line 115: please rename to onRemoveButtonClick

src/components/MaterialUI/DataDisplay/TransferList.js
```diff
@@ -0,0 +1,167 @@
...
+					</Button>
+					<Button
+						variant="outlined"
+						size="small"
+						className={classes.button}
+						onClick={onBottomButtonClick}
+						disabled={!rightSelected}
+						aria-label="move selected left"
+					>
+						&lt; {deleteButtonTitle || <FormattedMessage {...sharedMessages.remove} />}
```
InnaBoitsun, line 153: rename to removeButtonTitle

src/components/MaterialUI/DataDisplay/TransferList.js
```diff
@@ -0,0 +1,167 @@
...
+						disabled={!leftSelected}
+						aria-label="move selected right"
+					>
+						{addButtonTitle || <FormattedMessage {...sharedMessages.add} />} &gt;
+					</Button>
+					<Button
+						variant="outlined"
+						size="small"
+						className={classes.button}
+						onClick={onBottomButtonClick}
```
InnaBoitsun, line 149: use onRemoveButtonCLick

src/components/MaterialUI/DataDisplay/TransferList.js
```diff
@@ -0,0 +1,174 @@
...
+		onChange(formatState({ right: rightListData.data, left: leftListData.data }, checked, "right", "left"));
+		setChecked(checked.filter(check => !rightListData.data?.some(({ id }) => id === check)));
+	};
+
+	const rightSelected = rightListData.data?.some(({ id }) => checked.includes(id));
+	const leftSelected = leftListData.data?.some(({ id }) => checked.includes(id));
+
+	return (
+		<Grid container spacing={1} justify="center" alignItems="center" className={classes.root}>
+			<Grid item>
```
InnaBoitsun, line 130: Paper width should be responsive, I think Left and Right 5 cols, and for Buttons 2 cols

src/components/MaterialUI/DataDisplay/TransferList.js
```diff
@@ -0,0 +1,174 @@
...
+	},
+
+	title: {
+		color: theme.palette.text.hint,
+		fontSize: theme.spacing(1.2),
+		marginBottom: theme.spacing(0.6),
+	},
+	paper: {
+		overflow: "auto",
+		width: theme.spacing(32.6),
```
InnaBoitsun, line 26: width should be responside: left and right xs="5" , buttons = xs="2"

src/components/MaterialUI/DataDisplay/TransferList.js
```diff
@@ -0,0 +1,197 @@
...
+		lineHeight: theme.spacing(1.4),
+		padding: theme.spacing(0.7),
+		minWidth: theme.spacing(8),
+	},
+}));
+
+const formatState = (state, checked, from, to) => {
+	const fromSelectedItems = state[from].filter(({ id }) => checked.includes(id));
+	return {
+		[to]: [...state[to], ...fromSelectedItems].sort((a, b) => a.title.localeCompare(b.title)),
```
InnaBoitsun, line 69: I think we need to remove sort here - to able to have sort options from parent component which use this list. So we need to pass correct sorted items inside

src/components/MaterialUI/DataDisplay/TransferList.js
```diff
@@ -0,0 +1,197 @@
...
+					role="listitem"
+					onClick={!item.disabled ? () => handleToggle(item.id) : null}
+					disabled={item.disabled}
+				>
+					{listItemFormatter ? (
+						listItemFormatter(item)
+					) : (
+						<ListItemText
+							key={`transfer-list-item-${item.id}-label`}
+							primary={item.title}
```
InnaBoitsun, line 108: is it possible to make by default the stylw for primary title like:
fontWeight: theme.typography.fontWeightSemiBold

InnaBoitsun (dismissed):

src/components/MaterialUI/DataDisplay/TransferList.js
```diff
@@ -0,0 +1,203 @@
...
+	const onRemoveButtonClick = () => {
+		onChange(formatState({ right: rightListData.data, left: leftListData.data }, checked, "right", "left"));
+		setChecked(checked.filter(check => !rightListData.data?.some(({ id }) => id === check)));
+	};
+
+	const rightSelected = rightListData.data?.some(({ id }) => checked.includes(id));
+	const leftSelected = leftListData.data?.some(({ id }) => checked.includes(id));
+
+	return (
+		<Grid container spacing={1} justify="center" alignItems="center" className={classes.root}>
```
Crispried, line 148: Some time ago I realized that using spacing property on Grids without external wrapper (like div) could cause to some undesirable margins because of how that spacing property is implemented (it adds negative margin to a whole grid container).

So, what I would suggest is to wrap Grid container on the top in div just in case to avoid any issues in future.

src/components/MaterialUI/ScopeSelector/TreeView.js
```diff
@@ -3,49 +3,70 @@ import TreeViewMui from "@material-ui/lab/TreeView";
...
-		}
+	const onScopeSelect = (_, nodeId) => {
+		const newSelected = !Array.isArray(selected)
+			? nodeId
+			: selected.includes(nodeId)
+			? selected.filter(id => id !== nodeId)
+			: [nodeId, ...selected];
+
+		setSelected(newSelected);
+		closeSelector(_, newSelected);
```
Crispried, line 29: Usually parameter can/should be placeholder when we are not going to use it in handler. However the fact that now we are passing it to close selector means that we are started to use it in some way. So, now we need regular parameter (like event) instead of placeholder.

src/components/MaterialUI/ScopeSelector/TreeView.js
```diff
@@ -3,49 +3,70 @@ import TreeViewMui from "@material-ui/lab/TreeView";
...
 	};
 
 	const renderTree = scopeId => {
 		const currentScope = getScope(scopeId);
 		if (currentScope == null) return null;
 
-		if (currentScope.type === scopeTypes.virtual) {
-			virtualScopes.push(currentScope.id);
-		}
-
```
Crispried, line 29: So, since it's removed from there it should be handled somewhere else

Crispried: I didn't find anything what reimplements that behavior. Can you explain please?

oleksandr-honcharuk: is Virtual scope check already exist inside TreeItem, why need double check ?

InnaBoitsun (approved):

PR #210, opened 2021-02-24
updated table component

Anstsiya:
[AB#43803](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/43803)

### DIFF
--- a/src/components/MaterialUI/DataDisplay/Table.js
+++ b/src/components/MaterialUI/DataDisplay/Table.js
@@ -93,7 +93,7 @@ export const useStyles = makeStyles(theme => ({
 function headersAreIdentical(prevHeaders, nextHeaders) {
 	prevHeaders.forEach((prevHeader, index) => {
 		let prevSortOptions = prevHeader.cellElement.props.columnDefinition.sortOptions;
-		let nextSortOptions = nextHeaders[index].cellElement.props.columnDefinition.sortOptions;
+		let nextSortOptions = nextHeaders[index]?.cellElement.props.columnDefinition.sortOptions;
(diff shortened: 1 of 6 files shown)


src/components/MaterialUI/DataDisplay/Table.js
```diff
@@ -93,7 +93,7 @@ export const useStyles = makeStyles(theme => ({
 function headersAreIdentical(prevHeaders, nextHeaders) {
 	prevHeaders.forEach((prevHeader, index) => {
 		let prevSortOptions = prevHeader.cellElement.props.columnDefinition.sortOptions;
-		let nextSortOptions = nextHeaders[index].cellElement.props.columnDefinition.sortOptions;
+		let nextSortOptions = nextHeaders[index]?.cellElement.props.columnDefinition.sortOptions;
```
Crispried, line 96: I think we should just take sortOptions for "nextHeader" in the same way as we do it for "prevHeader"? So, nextHeaders.cellElement.props.columnDefintion.sortOptions.

Crispried (changes requested):

src/components/MaterialUI/DataDisplay/Table.js
Anstsiya: But it's array, We need index. And I added ? sign because in my story I have a case when in nextHeaders we have different amount of columns

Crispried: nvm, I missed the thing how you are comparing it

Crispried (approved):

PR #211, opened 2021-02-26
Add MUI wrapper for Timeline component

TONYFECTEAU:
Story ABC#41930


src/components/MaterialUI/DataDisplay/TimelineProps.js
```diff
@@ -0,0 +1,36 @@
+import ComponentProps from "../componentProps";
+import TimelineItemProps from "./TimelineItemProps";
+
+class TimelineProps extends ComponentProps {
+	static propNames = {
+		items: "items",
```
joclefrancois, line 6: "items" should not be a prop for a  TimeLine. A prop usually regards the behavior of the component, not its intrinsic nature. For example, rows are not a prop of the Table component, but its height could be,

src/components/MaterialUI/DataDisplay/TimelineItemProps.js
```diff
@@ -0,0 +1,27 @@
+import ComponentProps from "../componentProps";
+
+class TimelineItemProps extends ComponentProps {
+	static propNames = {
+		content: "content",
```
joclefrancois, line 5: Same thing for "content" and "contentOpposite"

src/components/MaterialUI/DataDisplay/Timeline.js
```diff
@@ -0,0 +1,71 @@
...
+import TimelineDotMui from "@material-ui/lab/TimelineDot";
+import TimelineProps, { isTimelineProps } from "./TimelineProps";
+import TimelineItemProps, { isTimelineItemProps } from "./TimelineItemProps";
+
+const renderItems = items => {
+	if (!items) return null;
+
+	let timelineItems = [];
+	let count = 0;
+	items.forEach(item => {
```
joclefrancois, line 16: You could have used map here, just saying.

joclefrancois (changes requested):
See my other comments

joclefrancois (approved):

PR #215, opened 2021-03-02
validation on require was implemented in edit state

Crispried:


src/constants.js
```diff
@@ -39,3 +39,9 @@ export const validationErrorTypes = {
 	fieldIsRequired: "fieldIsRequired",
 	fieldHasWrongValue: "fieldHasWrongValue",
 };
+
+export const mode = {
```
joclefrancois, line 43: mode of what? A more explicit name would be appropriate.

joclefrancois (approved):

src/constants.js
Crispried: done

joclefrancois (approved):

PR #219, opened 2021-03-04
Common styles for Modal Title:

InnaBoitsun:
@a-rousseau please review the common style for Modal Title. Right now we duplicate everywhere styles for title, let define some common one
image


a-rousseau (changes requested):
For the Modal we are missing the font color in the Title
**color: theme.palette.text.primary.main**

**LEAVE AS IS** (forget above comment)

Also make sure padding and margins for components are made in multiples of 10px like:
theme.spacing(0.5) around 5px
theme.spacing(1) around 10px
theme.spacing(2) around 20px

but visually the minimum should be 10px, except for buttons, dropdowns and similar components

a-rousseau (dismissed):

a-rousseau (approved):

PR #229, opened 2021-03-11
add able map model with nested fields

oleksandr-honcharuk:
[AB#41117](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/41117)

### DIFF
--- a/src/utils/mapHelper.js
+++ b/src/utils/mapHelper.js
@@ -22,7 +22,7 @@ export const mapModel = (model, initialModel, mappingRules = []) => {
 	const mapModifiedData = model => {
 		const mapModifiedObj = obj => {
 			if (typeof obj !== "object" || Array.isArray(obj)) return obj;
-			if (obj.wasModified !== undefined) return mapModifiedObj(obj.value);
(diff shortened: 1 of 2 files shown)


src/utils/mapHelper.js
```diff
@@ -19,22 +19,33 @@ export const mapModel = (model, initialModel, mappingRules = []) => {
 	// ]
 	// if you don't need this - just don't pass any value for that parameter
 
-	const result = initialModel != null ? cloneDeep(initialModel) : {};
+	const mapModifiedData = model => {
+		const mapModifiedObj = obj => {
+			if (typeof obj !== "object" || Array.isArray(obj)) return obj;
+			if (obj.wasModified !== undefined) return mapModifiedObj(obj.value);
```
Crispried, line 25: better to use != null instaed of !== undefined. In most of contexts null and undefined are pretty the same and you would like to handle both cases.

Crispried (approved):

Crispried (approved):

PR #242, opened 2021-03-17
fix issues with deep comparation

Anstsiya:
fixed bug [AB#44512](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/44512) , [AB#44562](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/44562) , fixed some design inconsistencies

### DIFF
--- a/src/components/MaterialUI/Inputs/Checkbox.js
+++ b/src/components/MaterialUI/Inputs/Checkbox.js
@@ -17,6 +17,7 @@ const useStyles = makeStyles(theme => ({
 			boxShadow: `0 0 4px ${theme.palette.focus}`,
 			outline: "none",
 		},
+		backgroundColor: theme.palette.background.default,
 	},
 	checkBoxIconChecked: {
 		backgroundColor: "currentColor",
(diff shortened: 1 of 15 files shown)


src/components/MaterialUI/Inputs/Checkbox.js
```diff
@@ -117,7 +118,7 @@ const Checkbox = ({ checkboxProps }) => {
 	const checkBoxMui = (
 		<CheckboxMUI
 			checked={value}
-			onChange={handleChange}
+			onChange={!readOnly ? handleChange : null}
```
InnaBoitsun, line 121: Does CheckboxMUI has disabled property? Beacuse it should be disabled for Readers (as we do for Inputs

Crispried: @InnaBoitsun 
We need to align readOnly and disabled props for most of input components with style guide create by Alain. Since currently in most of them we have just 1 but, need both.

We should do it as part of separate US and before it I'm suggesting to not care about it too much.

Crispried (approved):

src/components/MaterialUI/Inputs/Checkbox.js
InnaBoitsun: ok, but it is not hard to add disabled property here and use default MUI view for this. But as you wish.

Crispried (approved):

PR #253, opened 2021-03-24
Implement notification context and request processing engine

Crispried:
Notification context allows to wrap the app component in NotificationContextProvider and then access addNotification function through the NotificationContext. So, it makes notifications global and it won't be necessary to care anymore about component for notification. All what you need to do is just add new notification from anywhere and it will work!

RequestProcessingHelper exposes computeFinalRequestProcessingState which computes the final states depends on passed states. Using this function it will be possible to determine the global state of your "business action" which could contain multiple "http requests". 

It will be helpful also in context of **chain of request** which highly likely will be built in the future.

### DIFF
--- a/src/components/MaterialUI/Feedback/NotificationContext.js
+++ b/src/components/MaterialUI/Feedback/NotificationContext.js
@@ -1,0 +1,20 @@
+import React, { createContext, useState } from "react";
+import { useNotification } from "./useNotification";
+import Notification from "./../DataDisplay/Notification";
+
+export const NotificationContext = createContext();
+
+export const NotificationContextProvider = ({ children }) => {
+	const [snackPack, setSnackPack] = useState([]);
+
+	const addNotification = useNotification(snackPack, setSnackPack);
+
+	return (
+		<NotificationContext.Provider value={{ addNotification }}>
+			<Notification snackPack={snackPack} setSnackPack={setSnackPack} />
+			{children}
+		</NotificationContext.Provider>
+	);
+};
+
+export default NotificationContextProvider;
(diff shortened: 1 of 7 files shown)


src/components/MaterialUI/Feedback/NotificationContext.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const NotificationContext = createContext();
+
+export const NotificationContextProvider = ({ children }) => {
+	const [snackPack, setSnackPack] = useState([]);
+
+	const addNotification = useNotification(snackPack, setSnackPack);
+
+	return (
+		<NotificationContext.Provider value={{ addNotification }}>
```
TONYFECTEAU, line 13: Just curious, why the "."?

src/utils/requestProcessingHelper.js
```diff
@@ -0,0 +1,26 @@
...
+				return requestStates.fail;
+			case requestStates.processing:
+				return requestStates.processing;
+			case requestStates.idle:
+				idleCount++;
+				continue;
+			case requestStates.success:
+				continue;
+			default:
+				throw Error("You have to pass just requestStates type");
```
TONYFECTEAU, line 19: Only valid requestStates can be used

src/components/MaterialUI/Feedback/NotificationContext.js
Crispried: createContext creates object with 2 main properties - Provider and Consumer. Depends on your need you should use one of those 2

src/utils/requestProcessingHelper.js
Crispried: Is there really big difference between those 2 statements? 😄

PR #257, opened 2021-03-29
Ability to specify Fallback Value for Translations.

InnaBoitsun:
For example when Display Name is absent for Custom Attributes, we need to Display Attribute Name for Input Label instead.

### DIFF
--- a/src/utils/setTranslationWithFallback.js
+++ b/src/utils/setTranslationWithFallback.js
@@ -1,0 +1,13 @@
+import { flatten } from "./flatten";
+
+export const setTranslationWithFallbackField = (locale, obj, fallbackField, ...field) => {
+	if (!obj || !obj.getIn(flatten([field]))) return obj;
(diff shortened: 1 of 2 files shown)


src/utils/setTranslationWithFallback.js
```diff
@@ -0,0 +1,13 @@
+import { flatten } from "./flatten";
+
+export const setTranslationWithFallbackField = (locale, obj, fallbackField, ...field) => {
+	if (!obj || !obj.getIn(flatten([field]))) return obj;
+	const value = obj.getIn(flatten([field, locale])) || obj.get(fallbackField) || "";
+	return obj.setIn(flatten([field]), value);
```
oleksandr-honcharuk, line 6: looks like similar functions 

we can simplify this function
return setTranslationWithFallbackValue(locale, obj, obj?.get(fallbackField), ...field)

InnaBoitsun: done

oleksandr-honcharuk (approved):

PR #264, opened 2021-04-06
Update Buttons and DatePicker Styles

a-rousseau:
Add DatePicker disabled styling
Fixed some basic Button styling
Fixed some basic Select styling


orck-adrouin (approved):

src/components/MaterialUI/Inputs/DatePicker.js
```diff
@@ -79,7 +85,7 @@ const WrappedDatePicker = ({
 	};
 
 	return (
-		<label className={classes.datePickerWrapper}>
+		<label className={clsx(classes.datePickerWrapper, disabledCls)}>
```
joclefrancois, line 88: @Crispried In some places, we use "classNames" and some other, we use "clsx". Do we have any standard for that?

a-rousseau: @joclefrancois @Crispried  In this case clsx is used to display conditional/complex classes. For example if disabledCls is empty it will not be added to the className property.

joclefrancois (dismissed):

src/components/MaterialUI/Inputs/DatePicker.js
a-rousseau: after review in orc-shared, will be using classNames

joclefrancois (approved):

PR #268, opened 2021-04-07
Allow empty value for Date Picker  and add  new CheckBoxGroup Component

InnaBoitsun:
Allow empty value for Date Picker 
Add Error message to DatePicker
Add new CheckBoxGroup Component needed for Lookup Multiple values
image


src/utils/modelValidationHelper.js
```diff
@@ -1,9 +1,14 @@
-import { isStringNullOrWhitespace } from "./propertyValidator";
+import { isStringNullOrWhitespace, isString } from "./propertyValidator";
 import { validationErrorTypes } from "../constants";
 import { isEmpty } from "lodash";
 
 export const validationRules = {
-	[validationErrorTypes.fieldIsRequired]: value => (isStringNullOrWhitespace(value) ? false : true),
+	[validationErrorTypes.fieldIsRequired]: value => {
+		if (value === null || value === undefined) return false;
+		if (isString(value) && isStringNullOrWhitespace(value)) return false;
```
Crispried, line 8: I'm suggesting it make sense to add that isString check inside isStringNullOrWhitespace function. Since that function logically expecting string and if the value passed is not string then probably we would like to return true.

If you think it doesn't make sense feel free to ignore that suggestion.

src/components/MaterialUI/Inputs/CheckboxGroup.js
```diff
@@ -0,0 +1,72 @@
...
+	const update = checkboxGroupProps?.get(CheckboxGroupProps.propNames.update);
+	const value = checkboxGroupProps?.get(CheckboxGroupProps.propNames.value);
+	const label = checkboxGroupProps?.get(CheckboxGroupProps.propNames.label);
+	const error = checkboxGroupProps?.get(CheckboxGroupProps.propNames.error);
+	const readOnly = checkboxGroupProps?.get(CheckboxGroupProps.propNames.readOnly) || false;
+	const disabled = checkboxGroupProps?.get(CheckboxGroupProps.propNames.disabled) || false;
+	const options = checkboxGroupProps?.get(CheckboxGroupProps.propNames.options) ?? [];
+
+	const handleGroupUpdate = (checked, value, newValue) => {
+		let values = value ? value.split("|") : [];
```
Crispried, line 39: const values

src/components/MaterialUI/Inputs/CheckboxGroup.js
```diff
@@ -0,0 +1,72 @@
...
+		let position = values.indexOf(newValue);
```
Crispried, line 40: const position

src/components/MaterialUI/Inputs/CheckboxGroup.js
```diff
@@ -0,0 +1,72 @@
...
+			values.push(newValue);
+		}
+
+		if (!checked && position !== -1) {
+			values.splice(position, 1);
+		}
+		update(values.join("|"));
+	};
+
+	const classes = useStyles();
```
Crispried, line 51: better to place hook calls on the top of component so, by this way you will minimize chance that somebody won't notice it and add some conditional return before all hooks are called

Crispried (approved):

PR #277, opened 2021-04-12
Add confirmation dialog when changing scope

orck-adrouin:
#41583


TONYFECTEAU (approved):

src/components/Scope/useScopeConfirmationModalState.js
```diff
@@ -0,0 +1,71 @@
...
+		history.push(href);
+		dispatch(applicationScopeHasChanged(previousScope, newScope));
+	};
+};
+
+const useScopeConfirmationModalState = () => {
+	const [currentScope] = useScopeData();
+	const [isModalOpened, setModalOpened] = React.useState(false);
+	const hasUnsavedData = useSelector(hasUnsavedDataSelector);
+	const [newScope, setNewScope] = useState(null);
```
joclefrancois, line 35: Does it work? I had several problem using "useState" in a hook. My element, here "newScope" was always null event though I was calling setNewScope :-(

joclefrancois: Based on setModelOpened, I guess so

orck-adrouin: Yes it work well.

joclefrancois (approved):

PR #308, opened 2021-05-03
Add StepperModal type option (to allow wide)

TONYFECTEAU:
Add a new type option for StepperModal component
- Default will ne wide
- Allows normal, wide and fullwidth when using Modal component

Story ABCD#44268

### DIFF
--- a/src/components/MaterialUI/DataDisplay/PredefinedElements/StepperModal.js
+++ b/src/components/MaterialUI/DataDisplay/PredefinedElements/StepperModal.js
@@ -77,7 +77,15 @@ const useStyles = makeStyles(theme => ({
 	},
 }));
 
-const StepperModal = ({ steps = [], title = "", open, closeCallback, confirmCallback, confirmTitle }) => {
+const StepperModal = ({
+	steps = [],
+	title = "",
+	open,
+	closeCallback,
+	confirmCallback,
+	confirmTitle,
+	type = "normal",
+}) => {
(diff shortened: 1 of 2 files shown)


src/components/MaterialUI/DataDisplay/PredefinedElements/StepperModal.js
```diff
@@ -77,7 +77,15 @@ const useStyles = makeStyles(theme => ({
...
 
-const StepperModal = ({ steps = [], title = "", open, closeCallback, confirmCallback, confirmTitle }) => {
+const StepperModal = ({
+	steps = [],
+	title = "",
+	open,
+	closeCallback,
+	confirmCallback,
+	confirmTitle,
+	type = "normal",
```
a-rousseau, line 87: Default should be wide

a-rousseau (changes requested):

src/components/MaterialUI/DataDisplay/PredefinedElements/StepperModal.js
TONYFECTEAU: That will break Settings... they are using it for ManageRoleListModal component.

TONYFECTEAU: Changed to wide.

a-rousseau (dismissed):

PR #314, opened 2021-05-11
Update and fixes

a-rousseau:
+ Update Datepicker styling to timepicker layout
+ add title prop to InputBase for native tooltip
+ Update SearchControl to have clear button display conditionally if value is entered (through :valid css selector) and having it be a required field
+ Update SearchControl to be contained inside an actionless form to be able to use the reset functionality for the input
+ Fix SearchControl styling
+ Update button styling


src/components/MaterialUI/Inputs/PredefinedElements/SearchControl.js
```diff
@@ -131,13 +145,8 @@ const SearchControl = ({ placeholder, defaultValue = "", searchOptions, onSearch
...
 	};
 
-	const onClear = event => {
-		onSearch(searchOption, "");
-		inputRef.current.value = "";
-		inputRef.current.focus();
-
-		event.preventDefault();
-		event.stopPropagation();
+	const handleChange = e => {
```
TONYFECTEAU, line 148: Was this for debugging?  ;)

a-rousseau: this was the old code to clear the field. I used a new technique that removes the need for an onClick handler. A button of type="reset" resets the form without any extra code ! :)

TONYFECTEAU (dismissed):

TONYFECTEAU (approved):

PR #320, opened 2021-05-19
Fix issue when removing field from edit state always marks it as modified

Crispried:


src/reducers/view.js
```diff
@@ -53,7 +54,9 @@ const viewStateReducer = (state = initialState, action) => {
 
 			const pathToUpdate = newPath[newPath.length - 1] === "value" ? dropRight(newPath, 1) : newPath;
 
-			return state.setIn(pathToUpdate, Immutable.fromJS({ value, wasModified: true }));
+			const unpackedValue = mapModifiedData(value.toJS());
+
+			return state.setIn(pathToUpdate, Immutable.fromJS({ value, wasModified: !isEqual(unpackedValue, storeValue) }));
```
Anstsiya, line 59: Unfortunately it won't fix an issue fully. If storeValue is a full store but value from the newPath on level higher is not equal full store value, we still have issue

Crispried: changes applied

Anstsiya (approved):

PR #322, opened 2021-05-19
Add divider to left of info panel for SelectionList

TONYFECTEAU:
The SelectionList component will now put a divider property if it has any contents.

Story ABCD#44271

### DIFF
--- a/src/components/MaterialUI/DataDisplay/SelectionList.js
+++ b/src/components/MaterialUI/DataDisplay/SelectionList.js
@@ -2,6 +2,8 @@ import React, { useState } from "react";
 import { makeStyles } from "@material-ui/core/styles";
 import Grid from "@material-ui/core/Grid";
 import { ScrollableCustomList } from "./TransferList";
+import Divider from "./Divider";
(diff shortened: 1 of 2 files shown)


benaube (dismissed):

src/components/MaterialUI/DataDisplay/SelectionList.js
```diff
@@ -80,7 +88,12 @@ const SelectionList = ({
 					onChange={onChangeEvent}
 				/>
 			</Grid>
-			<Grid item xs={7}>
+			{infoPanel && (
+				<Grid item xs={1}>
+					<Divider dividerProps={dividerProps} />
+				</Grid>
+			)}
+			<Grid item xs={6}>
```
a-rousseau, line 96: should we have a condition if there is no infoPanel that this Grid item width is 7 instead of 6 ?

TONYFECTEAU: Makes sense, added it...  will push changes shortly.

a-rousseau (approved):

PR #330, opened 2021-05-25
Fix long texts in selector component

vadym-hyryn:


a-rousseau:
@vadym-hyryn  Is this fix for the Settings App ? 

Eventually this should be replaced in the code with the MaterialUI's version src/components/MaterialUI/Inputs/Select.js

a-rousseau (approved):

vadym-hyryn:
@a-rousseau  No, this is for the Analytics APP

PR #343, opened 2021-06-10
Features/add time zone translation

navidehsa:


src/components/MaterialUI/Inputs/DatePicker.js
```diff
@@ -116,7 +117,7 @@ const WrappedDatePicker = ({
 						onChange={date => updateDate(date, metadata)}
 						showTimeInput={useTime ?? false}
 						useTime={useTime ?? false}
-						customTimeInput={useTime ? <TimePicker showTimeZone={showTimeZone} /> : null}
+						customTimeInput={useTime ? <TimePicker showTimeZone={showTimeZone} timeZone={timePickerTimeZone} /> : null}
```
Anstsiya, line 120: don't you think we can pass here already translated timezone somehow? Cause we have all this timezones and translations in OMS for e.g. So maybe it's better than to have sooooo huge translations files with all timezones that some we even will never ever use

Anstsiya: actually getting timezones requests and selectors it's even a part of orc-shared. Can't we use just selector to get translation by timezone id ?

Anstsiya: I checked and it's possible to get timezones inside the timepicker.
I tried just to get all timezones like this 
image
and it works. So there should be no problems to pass timezone id and use localizedTimezone selector

navidehsa: @Anstsiya as far as I know the API does not return the translations and just return the English name,let me know if I am wrong!

Anstsiya (dismissed):

Anstsiya (approved):

Anstsiya (approved):

PR #346, opened 2021-06-10
Prompt the user that authentication needs to be refreshed

joclefrancois:
Prompt the user that authentication needs to be refreshed when cookies are expired.

Story: #47356


benaube (dismissed):

src/components/AppFrame/AppFrame.js
```diff
@@ -74,6 +94,12 @@ const AppFrame = ({ initOpen, applicationId, modules, activeModules, children, n
 	return (
 		<Base preferencesOpen={prefViewState.show}>
 			<ConnectedToastList />
+			<ActionModal
+				title={formatMessage(sharedMessages.error)}
+				message={formatMessage(sharedMessages.needToRefresh)}
```
a-rousseau, line 99: While you are in orc-shared I would like for you to add to ActionModal message a default styling. Currently the text is too big.  it should be using 

fontSize: theme.typography.fontsize,
fontFamily: theme.typography.fontFamily,

a-rousseau (approved):

PR #353, opened 2021-06-23
Allow autoComplete for InputBase to be set

TONYFECTEAU:
- Allow autoComplete for InputBase to be set

Story #48413

### DIFF
--- a/src/components/MaterialUI/Inputs/InputBase.js
+++ b/src/components/MaterialUI/Inputs/InputBase.js
@@ -98,6 +98,7 @@ const InputBase = ({ inputProps }) => {
 	const startAdornment = inputProps?.get(InputBaseProps.propNames.startAdornment);
 	const endAdornment = inputProps?.get(InputBaseProps.propNames.endAdornment);
 	const metadata = inputProps?.get(InputBaseProps.propNames.metadata);
+	const autoComplete = inputProps?.get(InputBaseProps.propNames.autoComplete);
 
 	const onClick = item => {
 		// Fixes FireFox issue, where the input number buttons do not focus on input control,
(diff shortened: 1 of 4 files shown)


benaube (approved):

src/components/MaterialUI/Inputs/InputBase.js
```diff
@@ -141,6 +142,7 @@ const InputBase = ({ inputProps }) => {
 					endAdornment={endAdornment}
 					rows={4}
 					title={value}
+					autoComplete={autoComplete === "off" ? "new-password" : autoComplete}
```
joclefrancois, line 145: So, there would be not way to turn off the "autoComplete" if someone wants to one day. Not sure it is a good idea.

src/components/MaterialUI/Inputs/InputBase.test.js
```diff
@@ -185,4 +185,23 @@ describe("InputBase Component", () => {
 
 		expect(mountedComponent.containsMatchingElement(expected), "to be truthy");
 	});
+
+	it("Renders InputBase with autoComplete", () => {
```
joclefrancois, line 189: Add another test with autoComplete having another value.

joclefrancois (changes requested):

src/components/MaterialUI/Inputs/InputBase.js
TONYFECTEAU: If autoComplete is null (the default) the default autoComplete behaviour is in place.  I tested this to make sure nothing changes unless you want to change it.  You can also set the autoComplete to the field that it will pull from, which is kind-of cool.

src/components/MaterialUI/Inputs/InputBase.test.js
TONYFECTEAU: Sure, will do that now.

PR #362, opened 2021-06-30
Resuming entity editing does not display modification indicators

joclefrancois:
- Resuming entity editing does not display modification indicators
- Run validations when required and when saving initial value in edit state
- Add ability to fully remove a portion of the edit state

Story: #46785

### DIFF
--- a/src/actions/view.js
+++ b/src/actions/view.js
@@ -1,12 +1,13 @@
 export const VIEW_SET = "VIEW_STATE_SET";
 export const VIEW_SET_FIELD = "VIEW_STATE_SET_FIELD";
 
-export const VIEW_CREATE_EDIT_NODE = "VIEW_CREATE_EDIT_NODE";
 export const VIEW_REMOVE_EDIT_NODE = "VIEW_REMOVE_EDIT_NODE";
 export const VIEW_SET_EDIT_MODEL_FIELD = "VIEW_SET_EDIT_MODEL_FIELD";
+export const VIEW_SET_FULL_ENTITY_EDIT_MODEL = "VIEW_SET_FULL_ENTITY_EDIT_MODEL";
 export const VIEW_SET_EDIT_MODEL_ERRORS = "VIEW_SET_EDIT_MODEL_ERRORS";
 export const VIEW_SET_EDIT_MODEL_FIELD_ERRORS = "VIEW_SET_EDIT_MODEL_FIELD_ERRORS";
 export const VIEW_REMOVE_EDIT_MODEL = "VIEW_REMOVE_EDIT_MODEL";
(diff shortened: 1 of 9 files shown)


TONYFECTEAU:
Missing code coverage on useFullEntityEditState.js

TONYFECTEAU (dismissed):

src/hooks/useEditState.js
```diff
@@ -22,46 +22,62 @@ export const useEditState = (entityId, sectionName, extendedValidationRules = {}
 
 	const mergedValidationRules = { ...validationRules, ...extendedValidationRules };
 
-	const useFieldState = (keys, initialValue = "", errorTypes = [], saveInitialValueToEditState = false) => {
+	const useFieldState = (
+		keys,
+		initialValue = "",
+		errorTypes = [],
+		saveInitialValueToEditState = false,
+		validateInitialValueAfterSave = false,
```
Crispried, line 30: Why you need those changes instead of just using dynamic edit state which allows you to do that? Technically dependencies means that there are dynamic relations so, it make sense to use dynamic edit state for that. Previously we already discussed it with team and it was decided to use dynamic edit state instead of modify static.
Regarding validateInitialValueAfterSave - I dont understand why you need that field? You can call isValid from the app to validate edit state when its necessary. Having this field makes that behavior redundant

src/hooks/useFullEntityEditState.js
```diff
@@ -0,0 +1,69 @@
+import Immutable from "immutable";
+import { useDispatchWithModulesData } from "./../hooks/useDispatchWithModulesData";
+import { setFullEntityEditModel } from "./../actions/view";
+import { validationRules } from "../utils/modelValidationHelper";
+import { isEqual } from "lodash";
+
+// if you need to override default validation rules just pass new rules with default keys
+// as properties of extendedValidationRules
+export const useFullEntityEditState = (
```
Crispried, line 9: May I ask you why you need that and cannot use dynamic edit state?

src/actions/view.js
```diff
@@ -35,6 +36,16 @@ export const setEditModelField = (keys, value, storeValue, entityId, sectionName
...
 });
 
+export const setFullEntityEditModel = (entityFullModel, moduleName) => ({
+	type: VIEW_SET_FULL_ENTITY_EDIT_MODEL,
+	payload: { entityFullModel, moduleName },
+});
+
+export const removeEditAllModelFields = (keys, entityId, sectionName, moduleName) => ({
+	type: VIEW_REMOVE_EDIT_ALL_MODEL_FIELDS,
+	payload: { keys, entityId, sectionName, moduleName },
```
Crispried, line 46: I believe we dont need that and mix of existing actions + dynamic edit state should allow you to work with any edit state

TONYFECTEAU (approved):

PR #387, opened 2021-07-21
Added error display to radiobuttons

orck-fsimard:


src/components/MaterialUI/Inputs/Radio.js
```diff
@@ -116,12 +125,22 @@ const Radio = ({ radioProps }) => {
 									onClick={handleClick}
 								/>
 							}
-						/>
+						></FormControlLabel>
```
TONYFECTEAU, line 128: Unless something is in the middle, can you just close the tag like it was before?

joclefrancois (approved):

src/components/MaterialUI/Inputs/Radio.js
```diff
@@ -116,12 +125,22 @@ const Radio = ({ radioProps }) => {
...
 					);
 				})}
 			</RadioGroupMui>
 		</FormControl>
 	);
+
+	return (
+		(error && (
```
TONYFECTEAU, line 136: Is this what we want?  To output the error if any instead of the radio button?

TONYFECTEAU (approved):

src/components/MaterialUI/Inputs/Radio.js
orck-fsimard: Fixed!

TONYFECTEAU (approved):

PR #394, opened 2021-07-28
Added error display to radiobuttons

orck-fsimard:


joclefrancois (changes requested):
Make sure the PR includes latest modifications from "releases/1.0.0", some modifications in the PR should not be in this Pull Request.

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "orc-shared",
-	"version": "1.0.0-pre.5",
+	"version": "1.0.0-pre.1",
```
TONYFECTEAU, line 3: Hmmm... Is it ok to use an older version here?

src/constants.js
```diff
@@ -40,7 +40,6 @@ export const validationErrorTypes = {
 	fieldHasWrongValue: "fieldHasWrongValue",
 	fieldMustBeValidEmail: "fieldMustBeValidEmail",
 	fieldMustBeValidPhoneNumber: "fieldMustBeValidPhoneNumber",
-	fieldMustBeValidPhoneExtension: "fieldMustBeValidPhoneExtension",
```
TONYFECTEAU, line 43: Are we removing Christopher's code?

TONYFECTEAU (approved):

PR #441, opened 2021-09-20
Add new html template icon

a-rousseau:
add html template icon in the iconsheet


src/content/iconsSheet.svg
```diff
@@ -312,6 +312,7 @@
...
 </symbol>
+<symbol id="icon-html-templates" /></symbol>
```
TONYFECTEAU, line 315: Am I missing something here? Shouldn't there be a path?

TONYFECTEAU (approved):

src/content/iconsSheet.svg
a-rousseau: that is created by the script ... this is weird ...

TONYFECTEAU (approved):

khanvong-ork (approved):

orck-adrouin (approved):

PR #466, opened 2022-03-23
SubPage - Render buttons passed from props instead of existing ones, if any

jsares:
Render buttons received from props if any. Render a single close button otherwise

### DIFF
--- a/src/components/Routing/SubPage.js
+++ b/src/components/Routing/SubPage.js
@@ -69,7 +69,10 @@ export const SubPage = ({ config, match, location, history, root, modulePrependP
 		</div>
 	);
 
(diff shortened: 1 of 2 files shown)


khanvong-ork (changes requested):
Code looks find. Need some more test coverage.
One thing is the arbitrary 200 timeout value. any particular reason we have a setTimeout and of 200 ?

jcmascolo-orckestra:
Looks fine

TONYFECTEAU (dismissed):

khanvong-ork (approved):
Approve just fix the coverage

khanvong-ork (approved):

PR #473, opened 2022-04-20
allow adaptive height when multiline and add max line count bypass

jsares:
- Allow adaptive height when multiline
- Add max line count bypass

 AB#60347

### DIFF
--- a/src/components/MaterialUI/DataDisplay/PredefinedElements/InformationItem.js
+++ b/src/components/MaterialUI/DataDisplay/PredefinedElements/InformationItem.js
@@ -53,7 +53,7 @@ const useStyles = makeStyles(theme => ({
 	},
 }));
 
-const InformationItemChildren = ({ classes, children, showNotAvailable }) => {
+const InformationItemChildren = ({ classes, children, showNotAvailable, isMaxLineCountEnabled = true }) => {
 	const { formatMessage } = useIntl();
 
 	if (isReactComponent(children)) {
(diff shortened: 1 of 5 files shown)


src/components/MaterialUI/DataDisplay/PredefinedElements/InformationItem.js
```diff
@@ -53,15 +53,15 @@ const useStyles = makeStyles(theme => ({
...
+const InformationItemChildren = ({ classes, children, showNotAvailable, isMaxLineCountEnabled = true }) => {
 	const { formatMessage } = useIntl();
 
 	if (isReactComponent(children)) {
 		return children;
 	}
 
 	const multipleLinesTextProps = new TextProps();
-	multipleLinesTextProps.set(TextProps.propNames.lineCount, 2);
+	if (isMaxLineCountEnabled) multipleLinesTextProps.set(TextProps.propNames.lineCount, 2);
```
TONYFECTEAU, line 64: I can see a future where we need to set the number of lines... this works for your PR, but maybe we should have a default max lines?

TONYFECTEAU (approved):

src/components/MaterialUI/DataDisplay/PredefinedElements/InformationItem.js
jsares: Interesting. Although I can see the ground work for that is already present, I would consider this as being a separate feature and that would have to be done in a separate PR.

PR #479, opened 2022-11-22
Allow validation before close Sub Page

InnaBoitsun:
[AB#55244](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/55244)

### DIFF
--- a/src/components/Routing/SubPage.js
+++ b/src/components/Routing/SubPage.js
@@ -81,7 +81,7 @@ export const SubPage = ({ config, match, location, history, root, modulePrependP
 						disabled={action.disabled}
 						onClick={e => {
 							let actionHandlerResult = action.handler && action.handler(e);
-							if (action.validateBeforeClose && !actionHandlerResult) return;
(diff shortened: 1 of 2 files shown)


src/components/Routing/SubPage.js
```diff
@@ -80,7 +80,8 @@ export const SubPage = ({ config, match, location, history, root, modulePrependP
 						disableElevation={action.isPrimary}
 						disabled={action.disabled}
 						onClick={e => {
-							action.handler && action.handler(e);
+							let actionHandlerResult = action.handler && action.handler(e);
+							if (action.validateBeforeClose && !actionHandlerResult) return;
```
neexite: Should actionHandlerResult be false when action.handler is null?

neexite (dismissed):

src/components/Routing/SubPage.js
InnaBoitsun: fixed the condition

neexite (approved):

PR #486, opened 2023-03-30
Integrate tasks related action/reducers and dialog

orck-adrouin:
#71230


TONYFECTEAU (approved):
Lots of changes for orc-shared.  It has been a while.

PR #519, opened 2023-12-01
Allow modal actions to be disabled

joclefrancois:
Story: [AB#73620](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/73620)


src/components/MaterialUI/DataDisplay/PredefinedElements/ActionModal.test.js
```diff
@@ -66,6 +66,68 @@ describe("ActionModal", () => {
 		expect(component, "when mounted", "to satisfy", expected);
 	});
 
+	it("Renders ActionModal correctly with disabled actions", () => {
```
orck-adrouin, line 76: Can you modify the previous test and add `disabled={false}` to the expected action?

orck-adrouin (approved):

PR #539, opened 2024-06-27
UI for inactive scope for Search  App

neexite:
Grey out inactive scopes for applications that supported inactive scopes.

[AB#77182](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/77182)

### DIFF
--- a/src/components/MaterialUI/ScopeSelector/TreeItem.js
+++ b/src/components/MaterialUI/ScopeSelector/TreeItem.js
@@ -34,6 +34,9 @@ const useStyles = makeStyles(theme => ({
 		display: "flex",
 		alignItems: "center",
 	},
+	scopeLabelInactive: {
+		color: theme.palette.secondary.light,
+	},
 	scopeIcon: {
 		marginRight: theme.spacing(0.7),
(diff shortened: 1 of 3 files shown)


src/components/MaterialUI/ScopeSelector/TreeItem.js
```diff
@@ -102,7 +105,7 @@ export const ScopeIcon = ({ type }) => {
 	}
 };
 
-export const ScopeLabel = ({ name, type, isRootScope, hasChildren, isVirtualScope }) => {
+export const ScopeLabel = ({ name, type, isRootScope, hasChildren, isVirtualScope, isActive }) => {
```
TONYFECTEAU: If you do not have a default for Is active you will need to change all references to this component. Basically breaking compatibility. Can you assign isactive as true to not break the component?

TONYFECTEAU: Maybe you can change the component doc as w well with this new attribute?

TONYFECTEAU (approved):

PR #546, opened 2024-07-18
Switch and DatePicker: use specific classes for readonly instead of props

Charles-Coque-Orckestra:
Switch and DatePicker: use specific classes for readonly instead of props

Bug: AB#78484

### DIFF
--- a/src/components/MaterialUI/Inputs/DatePicker.test.js
+++ b/src/components/MaterialUI/Inputs/DatePicker.test.js
@@ -3,8 +3,10 @@ import DatePicker, { createFormat } from "./DatePicker";
 import { mount } from "enzyme";
 import sinon from "sinon";
 import Icon from "../DataDisplay/Icon";
-import { TestWrapper, createMuiTheme } from "./../../../utils/testUtils";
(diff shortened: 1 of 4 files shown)


TONYFECTEAU (approved):

orck-adrouin (approved):

src/components/MaterialUI/Inputs/DatePicker.test.js
```diff
@@ -426,6 +428,38 @@ describe("DatePicker", () => {
...
+	});
+
+	it("should have updatable styles when updatable", () => {
+		const date = new Date("2020-06-30T00:00:00");
+
+		const component = (
+			<TestWrapper provider={{ store }} intlProvider>
+				<StylesProvider generateClassName={generateClassName}>
+					<MuiThemeProvider theme={createMuiTheme()}>
+						<DatePicker useTime={true} readOnly={true} onChange={updater} value={date} />
```
orck-adrouin: I believe readOnly=true should be false

Charles-Coque-Orckestra: Yes it should ! Fixed !

PR #567, opened 2025-03-05
Features/82083 add menu button

Charles-Coque-Orckestra:
[AB#82083](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/82083)


TONYFECTEAU (approved):
Can you add to docs/components md

src/components/MaterialUI/MenuButtons.js
```diff
@@ -0,0 +1,107 @@
...
+				aria-haspopup="true"
+				variant="outlined"
+				color="primary"
+				disabled={allDisabled}
+				onClick={handleClick}
+				endIcon={<Icon id="chevron-down" className={classes.arrowIcon} />}
+			>
+				{label}
+			</Button>
+			<StyledMenu id="actions-menu" anchorEl={anchorEl} open={Boolean(anchorEl)} onClose={handleClose}>
```
orck-adrouin: is this `id` property useful or end up in the DOM ?

Charles-Coque-Orckestra: No ! I used it during dev and did not realize i left it, i'll remove it

Charles-Coque-Orckestra: but now i realize it is linked with the aria-controls

Charles-Coque-Orckestra: I added a name property to identify the component, that way we can keep the aria attributes and avoid a duplication of the id if multiple menubutton are created in the same page

Wait i forgot a name property already exist and that's weird to use it as an id

TONYFECTEAU (approved):

orck-adrouin (approved):

PR #578, opened 2025-06-11
Features/82865 remove orc shared components

OlesOrckestra:
feature: [AB#82865](https://dev.azure.com/orckestra001/OrckestraCommerce/_workitems/edit/82865)

Main changes in this PR:
1. /src/components/Form/*
- removed not used components
- moved from styled-packages to material UI version. It's very specific for report builder in Analytics so it didn't make much sense to fully remove it and rewrite the reports builder. I removed the components which are not used in the report builder
2. /src/components/DropMenu/*
- moved from styled-packages to material UI version. Used in search and in Toolbar in AppFrame 
3.  input-related components in /src/components
- updated OMS, Settings, Analytics, and Search not to use some of the main components (I will create separate PRs in corresponding repositories): Text, Placeholder, ErrorPlaceholder, IconButton, ColumnWrapper
- I can't remove many of them yet as we still have to clean List, AppFrame, and other components first


joclefrancois (changes requested):
Hello Oles,

It is more than likely the user story might not have been clear enough, but we also want to remove components that have something similar in "MaterialUI/Inputs".

For example, we have Button or Text components in both places, we only want to keep the MaterialUI version and get rid of the old one. After that, we need to change Analytics, Settings and OMS to remove and replace references to old components.

You also added "Selector" but it uses our Select implementation based on MaterialUI. It would be preferable to use <Select/>directly.

Contact me when possible, probably next Tuesday.

Regards

OlesOrckestra:
> Hello Oles,
> 
> It is more than likely the user story might not have been clear enough, but we also want to remove components that have something similar in "MaterialUI/Inputs".
> 
> For example, we have Button or Text components in both places, we only want to keep the MaterialUI version and get rid of the old one. After that, we need to change Analytics, Settings and OMS to remove and replace references to old components.
> 
> You also added "Selector" but it uses our Select implementation based on MaterialUI. It would be preferable to use directly. Contact me when possible, probably next Tuesday. Regards

Hello Jocelyn,

Let's discuss it on a call today. Call me when you are available

joclefrancois (approved):

PR #580, opened 2025-06-19
Refactored Preferences and withScrollBox components

OlesOrckestra:
Story: [AB#82865](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/82865)

### DIFF
--- a/src/components/AppFrame/Preferences.js
+++ b/src/components/AppFrame/Preferences.js
@@ -5,8 +5,6 @@ import { getThemeProp, memoize, unwrapImmutable } from "../../utils";
 import Sidepanel from "../Sidepanel";
 import Button from "../Button";
 import Text from "../Text";
-import { FormContext } from "../Form/Form";
-import FieldElements from "../Form/FieldElements";
 import useViewState from "../../hooks/useViewState";
 import { setValue } from "../../actions/view";
 import { setDefaultLanguage } from "../../actions/locale";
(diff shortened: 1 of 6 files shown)


TONYFECTEAU (approved):

src/components/AppFrame/Preferences.js
```diff
@@ -29,6 +30,7 @@ export const PrefPanel = styled(Sidepanel)`
 	top: 40px;
 	box-shadow: -3px 2px 5px 0px rgba(0, 0, 0, 0.2);
 	z-index: 9998;
+	transition: right 400ms ease-out;
 `;
 
 export const Wrapper = withClickOutside(styled.div`
```
joclefrancois, line 36: Why "styled" is still used here?

src/hocs/withScrollBox.js
```diff
@@ -11,15 +11,20 @@ const useStyles = makeStyles({
 	},
 });
 
-const withScrollBox =
-	WrappedComp =>
-	({ onScroll, ...otherProps }) => {
+const withScrollBox = WrappedComp =>
+	React.forwardRef(({ onScroll, ...otherProps }, externalRef) => {
```
joclefrancois, line 15: Just curious what was the motivations behind this change?

src/components/AppFrame/Preferences.js
OlesOrckestra: We have a 5 story points story dedicated for AppFrame:
https://dev.azure.com/orckestra001/OrckestraCommerce/_workitems/edit/82864

I wanted to update all AppFrame components in this story

src/hocs/withScrollBox.js
OlesOrckestra: I need to pass the ref because it's used in some places in OMS:
image

Then when we add a new parameter, the ref is used:
image

joclefrancois: Ok, can you check with Alex about that please? I will approve your PR, but I know Alex and Charles worked on those scrollbars, they also discovered strange behavior in some place, I just want to make sure it won't interfere with their work.

joclefrancois (approved):

src/hocs/withScrollBox.js
orck-adrouin: the code that Charles modified was for the Table component so it should be OK.

PR #589, opened 2025-08-21
Features/82864 refactor app frame components

OlesOrckestra:
Refactored App Frame components

Story: [AB#82864](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/82864)

1. Updated each AppFrame component not to use "styled-components".
2. Updated each AppFrame component not to use old orc-shared components (Icon, Text, etc.).


TONYFECTEAU (approved):

src/components/AppFrame/AppFrame.js
```diff
@@ -121,7 +112,7 @@ const AppFrame = ({ initOpen, applicationId, modules, activeModules, children, n
 				onClick={reset}
 			/>
 			<Sidebar {...{ open, toggle, modules: menuItemFromModules, activeModules }} />
```
joclefrancois, line 114: You don't use "open" anymore?

src/components/AppFrame/MenuItem.js
```diff
@@ -1,147 +1,127 @@
...
-`;
-
-const MenuItem = ({ open = false, label = "", icon, alert, isHidden = false, href, ...props }) => {
-	let ItemWrapper = Block;
-	if (props.menuToggle) {
-		ItemWrapper = BlockWithA;
-	}
 	const alertMessage = useRef("");
+	const showAlert = !!alert;
+	const showAlertMessage = !!alert?.message;
```
joclefrancois, line 106: Should you not use those flags now in the "if" below?

src/components/AppFrame/MenuItem.js
```diff
@@ -1,147 +1,127 @@
...
 	if (alert && alert.message) {
 		alertMessage.current = alert.message;
 	}
 
+	if (isHidden) {
+		return null;
```
joclefrancois: Just a nice to know, if you are interested and if you did not know already:

https://medium.com/@davidkelley87/stop-using-return-null-in-react-a2ebf08fc9cd

src/components/AppFrame/MenuItem.js
```diff
@@ -1,147 +1,127 @@
...
-					</Alert>
-				) : null}
-				<Label show={open}>
-					<Text message={label} />
-				</Label>
-			</ItemWrapper>
-		)
+		<ItemWrapper to={href} className={classes.block} {...props}>
+			<Icon id={icon} className={classes.menuIcon} />
+			<div className={`${classes.alert} ${showAlert ? "show" : undefined}`}>
```
joclefrancois, line 120: Why do you still render the <div/> section if "showAlert" is false, I definitely prefer the old syntax. I understand that without "show" class, if would not be visible, yet, the <div/> would still be renderer.

Only my opinion, but generally, I prefer the old syntax.

OlesOrckestra: I was still rendering these elements to support the existing CSS animation.

It's not ideal since we have to render the elements as you said. Since we mostly used "styled-transition-group", I didn't see any better approaches we are using.

I can rewrite using react-transition-group. Do you have any other options we can use for the new animations?

src/components/AppFrame/AppFrame.js
OlesOrckestra: I moved the "open" to the props:
const classes = useStyles({ preferencesOpen: prefViewState.show, toggleOpen: open });
width: props.toggleOpen ? "calc(100% - 200px)" : "calc(100% - 50px)",

src/components/AppFrame/MenuItem.js
joclefrancois: I would need to see the difference and I am not sure about the animation you are talking about, even in the old version.

PR #591, opened 2025-09-05
Added tooltip to information item

OlesOrckestra:
Added tooltip to the information item

Story: [AB#83341](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/83341)

### DIFF
--- a/src/components/MaterialUI/DataDisplay/PredefinedElements/InformationItem.js
+++ b/src/components/MaterialUI/DataDisplay/PredefinedElements/InformationItem.js
@@ -2,12 +2,14 @@ import React from "react";
 import { FormattedMessage } from "react-intl";
 import Typography from "@material-ui/core/Typography";
 import { makeStyles } from "@material-ui/core/styles";
+import Grid from "@material-ui/core/Grid";
 import MultipleLinesText from "../TooltippedElements/MultipleLinesText";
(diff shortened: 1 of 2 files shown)


src/components/MaterialUI/DataDisplay/PredefinedElements/InformationItem.js
```diff
@@ -100,6 +115,19 @@ const InformationItemHeader = ({ classes, label, headerIcon, headerIconClassName
...
+				<TooltippedIcon
+					titleValue={headerTooltip}
+					id={headerTooltipIcon}
+					className={classNames(classes.headerTooltip, headerTooltipClassName)}
+				/>
+			</Grid>
+		);
+	}
+
 	if (headerIcon) {
```
orck-adrouin, line 131: Can we use the `headerIcon` property to display the tooltip where needed instead of creating new tooltip properties? You can look at the warning icon in the order fulfillment section for a similar use case. `headerIcon` is not just an icon it can be any control that we want.


Image

OlesOrckestra: Thanks! We can pass not only icon, but any component so it works as well

Let's abandon this PR

PR #592, opened 2025-09-05
Fixing datePicker and timePicker based on culture and language

joclefrancois:
- Fixing datePicker and timePicker based on culture and language
- Defined an updated time zone mapping between IANA and Windows time zone definitions

Story: [AB#84550](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/84550)


src/components/MaterialUI/Inputs/TimePicker.js
```diff
@@ -65,8 +66,8 @@ const ampmOptions = [
 	{ value: "PM", label: "PM" },
 ];
 
-const isBrowserUsingAMPM = () =>
-	!!new Date(Date.UTC(2020, 7, 30, 3, 0, 0)).toLocaleTimeString().match(/am|a.m|pm|p.m/i);
+const isBrowserUsingAMPM = locale =>
+	!!new Date(Date.UTC(2020, 7, 30, 3, 0, 0)).toLocaleTimeString(locale).match(/am|a.m|pm|p.m/i);
```
orck-adrouin, line 70: this regex has a bug, it will match `aWm` and not just `a.m` (I am assuming we want to match the string `a.m.`)

Switch to /am|a\.m|pm|p\.m/i

orck-adrouin (approved):

src/components/MaterialUI/Inputs/TimePicker.js
joclefrancois: I must be crazy, but it seems your suggestions is identical to what we have currently. I will still complete my PR but will fix it next week if we still need to modify something.

PR #606, opened 2026-01-20
Conversion of px to theme.spacing()

joclefrancois:
Story: AB#82868


src/components/AppFrame/Preferences.js
```diff
@@ -139,7 +139,7 @@ export const Preferences = () => {
 	const { show, save, clear, languageOptions, languageSelectProps, applicationOptions, applicationSelectProps } =
 		usePreferenceSetup();
 	return (
-		<Sidepanel in={show} width="380px" timeout={400} className={classes.prefPanel}>
+		<Sidepanel in={show} widthSpacing={38} timeout={400} className={classes.prefPanel}>
```
orck-adrouin, line 142: should we put this value inside `prefPanel`?

src/components/DropMenu/Menu.js
```diff
@@ -7,7 +7,7 @@ const useStyles = makeStyles(theme => ({
 	drawer: {
 		position: "absolute",
 		zIndex: 19999,
-		margin: "4px 0 0",
+		margin: theme.spacing(0.4, 0, 0),
```
orck-adrouin, line 10: are we missing a number in the spacing method? Does it work with only 3 ?

src/components/Form/Field.js
```diff
@@ -13,7 +13,7 @@ const useStyles = makeStyles(theme => ({
 	label: {
 		minHeight: 17,
 		color: theme.palette.text.hint,
-		marginBottom: "10px",
+		marginBottom: theme.spacing(1),
```
orck-adrouin, line 16: we're missing the units for minHeight:17 and marginTop:20

src/components/MaterialUI/DataDisplay/PredefinedElements/Placeholder.js
```diff
@@ -51,7 +51,7 @@ const useStyles = makeStyles(theme => ({
 		width: 60,
 		height: 25,
 		animation: false,
-		borderRadius: "15px",
+		borderRadius: theme.spacing(1.5),
```
orck-adrouin, line 54: same for this file (missing units)

src/components/DropMenu/Menu.js
joclefrancois: The number of outputs returned by the spacing method is the same as the number of inputs, which ought to be between 0 and 4 arguments (don't ask me what is the use case for zero argument).

src/components/Scope/ScopeNode.js
```diff
@@ -5,10 +5,10 @@ import classNames from "classnames";
 
 const useStyles = makeStyles(theme => ({
 	scopeIcon: {
-		fontSize: "20px",
-		verticalAlign: "middle",
-		paddingRight: "8px",
-		flexShrink: 0,
+		paddingRight: theme.spacing(0.8),
```
orck-adrouin, line 8: wrong values or did you need to adjust the spacing?

orck-adrouin (approved):

src/components/Scope/ScopeNode.js
joclefrancois: I redefined completely the scope icon class here, most of it was useless.

src/components/AppFrame/Preferences.js
joclefrancois: Yes

PR #609, opened 2026-02-25
Improved new components in replacement of styled-components

joclefrancois:
Story: [AB#86373](https://dev.azure.com/orckestra001/da965986-c85b-43be-9915-219568b000e5/_workitems/edit/86373)


src/components/MaterialUI/Inputs/Select.js
```diff
@@ -171,139 +185,164 @@ const Select = ({ options, selectProps, children }) => {
...
+
+	const multipleSelectWidthFactor = windowSize.innerWidth > 1400 ? 1 : 0.75;
+	const classes = useStyles({
+		multiple,
+		autoWidth,
+		multipleWidthSpacing: (multipleSelectWidth / 10) * multipleSelectWidthFactor,
+	});
+
+	const buildOptionsItems = () => {
+		const allOptions = showAllValue && showAllLabel ? [...options] : options;
```
orck-adrouin: I'm not sure why we need to copy the array if both variables are true

src/components/MaterialUI/Inputs/Select.js
```diff
@@ -171,139 +185,164 @@ const Select = ({ options, selectProps, children }) => {
...
 
-	if (showAllValue && showAllLabel) {
-		options.unshift({
-			value: showAllValue,
-			label: showAllLabel,
+		if (sortType === sortTypeEnum.numeric) {
+			allOptions.sort((a, b) =>
+				a.sortOrder.localeCompare(b.sortOrder, undefined, {
```
orck-adrouin, line 217: do we always have a sortOrder property?

src/components/MaterialUI/Inputs/Select.js
```diff
@@ -171,139 +185,164 @@ const Select = ({ options, selectProps, children }) => {
...
+		}
+
+		if (showAllValue && showAllLabel) {
+			allOptions.unshift({
+				value: showAllValue,
+				label: showAllLabel,
+			});
+		}
+
+		return allOptions?.map(option => {
```
orck-adrouin: is it possible for `allOptions` to be null here?

src/components/MaterialUI/Inputs/Select.js
```diff
@@ -171,139 +185,164 @@ const Select = ({ options, selectProps, children }) => {
...
+			let clss = option?.level ? classes["level" + option.level] : "";
```
orck-adrouin: let => const

orck-adrouin (approved):

src/components/MaterialUI/Inputs/Select.js
joclefrancois: (Misread your question... cause, end of the day)

No, we don't, I left the code as it was before, even though I moved it in a function.

src/components/MaterialUI/Inputs/Select.js
joclefrancois: Because, if the select component is re-rendered, on 'resize' for example, we don't want to modify the original array again and again with this code you can see further down: 


allOptions.unshift({
				value: showAllValue,
				label: showAllLabel,
			});

I had some issues with unit tests because of that.

src/components/MaterialUI/Inputs/Select.js
joclefrancois: No, but with the old code, it was.

